PR #1, opened 2017-02-22
Add needed dependencies

aaronraimist:
After running `npm install` I got these warnings. I assume you just forgot to add them to the dependencies list.
```
npm WARN discord.js@11.0.0 requires a peer of erlpack@hammerandchisel/erlpack#master but none was installed.
npm WARN discord.js@11.0.0 requires a peer of node-opus@^0.2.0 but none was installed.
npm WARN discord.js@11.0.0 requires a peer of opusscript@^0.0.1 but none was installed.
npm WARN discord.js@11.0.0 requires a peer of uws@^0.12.0 but none was installed.
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,13 +32,17 @@
     "@types/node": "^7.0.5",
     "bluebird": "^3.4.7",
     "discord.js": "^11.0.0",
+    "erlpack": "hammerandchisel/erlpack#master",
     "js-yaml": "^3.8.1",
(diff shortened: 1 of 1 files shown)


Half-Shot:
You'd think but actually this is just for voice decoding/encoding, which is far off in the future. The warnings are a bit naff, but it's that or bundling some useless stuff.

PR #8, opened 2017-05-13
Set display name and avatar on initial join and handle joining/leaving guild members

Johni0702:
Previously the avatar and user name of a bridged discord user were only set on their matrix user after the first message.
Per guild display names were only set after the discord user changed their display name.

The first commit changes the initial auto-joining of users into a new matrix room to also set their avatar, user name and display name.

The second commit handles discord users that join/leave guilds by joining/parting their matrix counterparts into/from the respective matrix rooms.
In its current form, this will probably break if multiple discord channels were to be bridged to the same matrix channel; not sure if this configuration should be supported.

### DIFF
--- a/src/bot.ts
+++ b/src/bot.ts
@@ -229,6 +229,15 @@ export class DiscordBot {
     });
   }
 
+  public InitJoinUser(member: Discord.GuildMember, roomId: string): Promise<any> {
+    const intent = this.bridge.getIntentFromLocalpart(`_discord_${member.id}`);
+    return this.UpdateUser(member.user).then(() => {
+      return intent.join(roomId);
+    }).then(() => {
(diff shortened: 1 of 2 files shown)


turt2live:
This has a relatively easy merge conflict to resolve, but otherwise seems to work as intended (running in production).

strk:
@Johni0702 are you planning to rebase/fix-conflicts ?
Or @turt2live maybe you can push your own branch ?

Johni0702:
Rebased and added a missing semicolon.

strk:
Thanks @Johni0702 ! @Half-Shot anything we can do here to speed up merging ?

turt2live:
The original should be on my branch. I can't review the changes right now,
but if they're significant then I'll update.

sent from mobile

PR #9, opened 2017-05-23
Added initial support for webhooks. Need next version of discord.js b…

ollieh:
…efore file attachments will work through webhooks.

### DIFF
--- a/src/bot.ts
+++ b/src/bot.ts
@@ -159,7 +159,11 @@ export class DiscordBot {
     const profile = result.botUser ? await mxClient.getProfileInfo(event.sender) : null;
     const embed = this.MatrixEventToEmbed(event, profile, chan);
     let opts : Discord.MessageOptions = {};
-    if (["m.image", "m.audio", "m.video", "m.file"].indexOf(event.content.msgtype) !== -1) {
+    let hookOpts : Discord.WebhookMessageOptions = {
+      username: event.sender,
+    };
+    const hasAttachment = ["m.image", "m.audio", "m.video", "m.file"].indexOf(event.content.msgtype) !== -1;
+    if (hasAttachment) {
       const attachment = await Util.DownloadFile(mxClient.mxcUrlToHttp(event.content.url));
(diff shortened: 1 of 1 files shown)


ollieh:
Just realised if you have multiple rooms on the same discord server that you need to bridge, it won't work as it doesn't filter the hooks by room. I'll fix that.

src/bot.ts
```diff
@@ -168,12 +172,21 @@ export class DiscordBot {
...
+    const webhooks = await chan.fetchWebhooks();
+    const hook : Discord.Webhook = webhooks.filterArray((h) => h.name === "_matrix").pop();
     try {
-      if (botUser) {
+      if (!botUser) {
+        msg = await chan.send(embed.description, opts);
+      } else if (hook && !hasAttachment) { //Remove !hasAttachment when https://github.com/hydrabolt/discord.js/pull/1449 is fixed
+        if (hasAttachment) {
+          hookOpts.file = opts.file;
+          msg = await hook.send(embed.description, hookOpts);
```
Half-Shot: This should be the bot user, no?

src/bot.ts
```diff
@@ -159,7 +159,11 @@ export class DiscordBot {
     const profile = result.botUser ? await mxClient.getProfileInfo(event.sender) : null;
     const embed = this.MatrixEventToEmbed(event, profile, chan);
     let opts : Discord.MessageOptions = {};
-    if (["m.image", "m.audio", "m.video", "m.file"].indexOf(event.content.msgtype) !== -1) {
+    let hookOpts : Discord.WebhookMessageOptions = {
+      username: event.sender,
```
Half-Shot: Embed will contain a display name and a avatar, so you could put this in here.

src/bot.ts
Half-Shot: Looking at it, probably best to comment this out since the else aught to take care of this anyway

Half-Shot:
LGTM

PR #10, opened 2017-05-24
Add tests for matrix -> discord message sending and add permission to bot

ollieh:


src/bot.ts
```diff
@@ -39,7 +39,7 @@ export class DiscordBot {
     this.bridge = bridge;
   }
 
-  get ClientFactory(): DiscordClientFactory {
+  public getClientFactory(): DiscordClientFactory {
```
Half-Shot: What's the meaning of this change? I prefer to use the native get/set methods over pure functions?

Half-Shot:
Looks generally quite good, but needs the tests passing before I can merge :)

Half-Shot:
I'm going to close this since it's been inactive and conflicting with quite a few changes now.

PR #25, opened 2017-07-23
Random attempt at updating profile before joining

### LINKED ISSUE #24: Update profile before joining room
For the somewhat rare case of people bridging `irc <--> matrix <--> discord`. When the discord bridge joins the user to the matrix room it doesn't always have a display name set, so the IRC bridge picks up the user ID, leading to IRC users not having a clue who is speaking.

Blocking the join on a successful profile update would seem reasonable to me.

strk:
Hopefully fixes #24 (untested!)

\cc @TravisR (can you test this?)

### DIFF
--- a/src/discordbot.ts
+++ b/src/discordbot.ts
@@ -89,10 +89,7 @@ export class DiscordBot {
     let chan;
     let embed;
     const mxClient = this.bridge.getClientFactory().getClientAs();
-    return this.LookupRoom(guildId, channelId).then((channel) => {
-      chan = channel;
-      return mxClient.getProfileInfo(event.sender);
-    }).then((profile) => {
+    return this.LookupRoom(guildId, channelId).then((profile) => {
       if (!profile.displayname) {
         profile.displayname = event.sender;
       }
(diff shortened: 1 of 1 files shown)


src/discordbot.ts
```diff
@@ -89,10 +89,7 @@ export class DiscordBot {
     let chan;
     let embed;
     const mxClient = this.bridge.getClientFactory().getClientAs();
-    return this.LookupRoom(guildId, channelId).then((channel) => {
-      chan = channel;
-      return mxClient.getProfileInfo(event.sender);
-    }).then((profile) => {
+    return this.LookupRoom(guildId, channelId).then((profile) => {
```
turt2live: The variables don't match here. `LookupRoom` resolves to a channel, not a profile.

strk:
Is d231d99 any better ?

src/discordbot.ts
```diff
@@ -89,10 +89,7 @@ export class DiscordBot {
     let chan;
     let embed;
     const mxClient = this.bridge.getClientFactory().getClientAs();
-    return this.LookupRoom(guildId, channelId).then((channel) => {
```
turt2live: I think this diff is actually in the wrong spot now that I've looked a bit closer at it. This is on the matrix -> discord side, not the discord -> matrix side (the flow causing issues).

strk:
I hope you can look at this because I'm really just shooting in the
dark

turt2live:
Looks like the fix has been PRd already: https://github.com/Half-Shot/matrix-appservice-discord/pull/8

strk:
Closing this one then. Thanks for the heads up !

Issue #22, opened 2017-07-20
Multiple runs of npm start break database


anoadragon453:
Running `npm start` initially works fine and the bot starts up. However killing the process and running `npm start` again causes the following error output, forcing a full recreation of the database to start correctly again:

```
WARN MatrixRoomHandler Skipping event due to age 13307669916 > 900000 
WARN MatrixRoomHandler Skipping event due to age 323627349 > 900000 
 
^C 
root@ubuntu:~/.matrix-appservice-discord# npm start 
 
> matrix-appservice-discord@0.0.1 start /root/.matrix-appservice-discord 
> npm run-script build && node ./build/src/discordas.js -p 9005 -c config.yaml 
 
 
> matrix-appservice-discord@0.0.1 build /root/.matrix-appservice-discord 
> tsc 
 
Loading config file /root/.matrix-appservice-discord/config.yaml 
ERR! DiscordStore Couldn't update database to schema 2 
ERR! DiscordStore Error: Error creating 'dm_rooms': Error: SQLITE_ERROR: table dm_rooms already exists 
ERR! DiscordStore     at db.runAsync.then.catch (/root/.matrix-appservice-discord/build/src/store.js:96:19) 
ERR! DiscordStore     at tryCatcher (/root/.matrix-appservice-discord/node_modules/bluebird/js/release/util.js:16:23) 
ERR! DiscordStore     at Promise._settlePromiseFromHandler (/root/.matrix-appservice-discord/node_modules/bluebird/js/release/promise.js:512:31) 
ERR! DiscordStore     at Promise._settlePromise (/root/.matrix-appservice-discord/node_modules/bluebird/js/release/promise.js:569:18) 
ERR! DiscordStore     at Promise._settlePromise0 (/root/.matrix-appservice-discord/node_modules/bluebird/js/release/promise.js:614:10) 
ERR! DiscordStore     at Promise._settlePromises (/root/.matrix-appservice-discord/node_modules/bluebird/js/release/promise.js:689:18) 
ERR! DiscordStore     at Async._drainQueue (/root/.matrix-appservice-discord/node_modules/bluebird/js/release/async.js:133:16) 
ERR! DiscordStore     at Async._drainQueues (/root/.matrix-appservice-discord/node_modules/bluebird/js/release/async.js:143:10) 
ERR! DiscordStore     at Immediate.Async.drainQueues (/root/.matrix-appservice-discord/node_modules/bluebird/js/release/async.js:17:14) 
[...]

PR #34, opened 2017-08-24
Fix schema version update failure

Johni0702:
## Issue
On fresh installs, starting the appservice after the initial run, (without first removing the discord.db) causes the db upgrade to fail because some table already exists.

## Cause
The loop in `store#init(number)` upgrades the current db schema one version at a time. After each iteration, it updates the current schema version stored in the db. This update query is restricted to only update rows where the `version` column matches the old schema version.

This is fine for the first call (`this.setSchemaVersion(0, 1)`) but does nothing for the second call (`this.setSchemaVersion(0, 2)`) because the `oldVersion` in the loop never changes while the version in the db has changed.
This results in the final stored schema version to be `1` even though the db was actually successfully upgraded to `3` (or whatever the most recent one is). On the next startup, the upgrade from `1` is re-applied which fails as expected.

## Fix
One solution would have been to increment the `oldVersion` on every iteration of the loop (or just replace it by `version - 1`).
However, I couldn't see any good reason for keeping the `oldVersion` around in the first place, so this PR just removes it altogether.

### DIFF
--- a/src/store.ts
+++ b/src/store.ts
@@ -55,8 +55,7 @@ export class DiscordStore {
   public async init (overrideSchema: number = 0) {
     log.info("DiscordStore", "Starting DB Init");
     await this.open_database();
-    const oldVersion = await this.getSchemaVersion();
-    let version = oldVersion;
+    let version = await this.getSchemaVersion();
     const targetSchema = overrideSchema || CURRENT_SCHEMA;
     while (version < targetSchema) {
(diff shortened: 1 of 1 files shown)


anoadragon453:
Can confirm this fixes #22!!

Half-Shot:
Wow, it was that easy? Great. I'll merge it :+1:

PR #39, opened 2017-09-12
Add config option to specify default directory listing

### LINKED ISSUE #38: Config option to make new rooms non-public by default
Users of my bridge can connect to any discord channel, and it'll display on the public room directory.

...This might not be ideal, for reasons.

The ability to tell the bot to mark new rooms as not public (specifically to set [m.room.join_rules](https://matrix.org/docs/spec/client_server/r0.2.0.html#m-room-join-rules) to 'invite' instead of 'public'), would be really helpful!

anoadragon453:
This commit adds a new section and option in `config.yaml` for rooms and whether
they should be shown in the HS' public directory listing by default.

The default behavior is to show them.

Fixes #38.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -5,6 +5,7 @@ export class DiscordBridgeConfig {
   public auth: DiscordBridgeConfigAuth;
   public logging: DiscordBridgeConfigLogging;
   public database: DiscordBridgeConfigDatabase;
+  public room: DiscordBridgeConfigRoom;
 }
 
 class DiscordBridgeConfigBridge {
@@ -23,4 +24,8 @@ export class DiscordBridgeConfigAuth {
 }
 class DiscordBridgeConfigLogging {
   public level: string;
+}
+
+class DiscordBridgeConfigRoom {
+  public defaultVisibility: string;
 }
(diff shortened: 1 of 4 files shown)


anoadragon453:
@Half-Shot Should these options be added to `config.schema.yaml` as well?

Half-Shot:
@anoadragon453 Yep, otherwise the config might not validate.

src/config.ts
```diff
@@ -24,3 +25,7 @@ export class DiscordBridgeConfigAuth {
 class DiscordBridgeConfigLogging {
   public level: string;
 }
+
+class DiscordBridgeConfigRoom {
+  public defaultPublic: boolean;
```
Half-Shot: I'd prefer this option to be a string and we can just ask the user to specify public or invite, to make this more explicit.

anoadragon453: Yeah, good idea.

anoadragon453:
Changed to `string`, added to schema config, and moved `defaultPublic` under the new `room:` property in the sample config.

Half-Shot:
Only thing I'd suggest is calling it defaultVisibility rather defaultPublic. Otherwise lgtm

anoadragon453:
Changed.

Half-Shot:
LGTM :+1:  Thanks for the PR :)

PR #40, opened 2017-09-12
Add config options to disable typing notifications and presence.

turt2live:
These are negative flags so that existing configurations can still work without modification. Otherwise the fields would be required and potentially break existing installs.


Half-Shot:
LGTM :+1:

PR #49, opened 2017-09-18
Auto-create _matrix webhook in channel

anoadragon453:
Creates the _matrix webhook if it doesn't exist already.

### DIFF
--- a/src/bot.ts
+++ b/src/bot.ts
@@ -239,9 +239,10 @@ export class DiscordBot {
     log.verbose("DiscordBot", `Event:`, event);
     const storeEvent = await this.store.Get(DbEvent, {matrix_id: event.redacts + ";" + event.room_id});
     if (!storeEvent.Result) {
-        log.warn("DiscordBot", `Could not redact because the event was in the store.`);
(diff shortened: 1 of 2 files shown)


src/bot.ts
```diff
@@ -194,6 +194,10 @@ export class DiscordBot {
     if (botUser) {
       const webhooks = await chan.fetchWebhooks();
       hook = webhooks.filterArray((h) => h.name === "_matrix").pop();
+      // Create a new webhook if none already exists
+      if (!hook) {
+        hook = await chan.createWebhook("_matrix", "", "Matrix Bridge: Allow rich user messages");
```
Half-Shot: It would probably be good to wrap this in a try catch with some logging, since this would be a prime spot for permission failures. Gold star if you can check the bots permissions before trying to create the webhook also since that isn't guaranteed.

anoadragon453: Presumably we'll just get a permissions error if we don't have permission when creating the webhook. Would is thus be enough to just print out the error and move along or is it possible to request the permission somehow?

TL;DR do we need to check if trying to create it will tell us anyways.

Half-Shot:
Yeah, this all looks fine. Merging.

PR #50, opened 2017-09-18
Add config option to disable message deletion forwarding

anoadragon453:


src/bot.ts
```diff
@@ -256,11 +258,13 @@ export class DiscordBot {
         const channel = <Discord.TextChannel> this.bot.guilds.get(storeEvent.GuildId)
                         .channels.get(storeEvent.ChannelId);
         const msg = await channel.fetchMessage(storeEvent.DiscordId);
-        try {
-            await msg.delete();
-            log.info("DiscordBot", `Deleted message`);
-        } catch (ex) {
-            log.warn("DiscordBot", `Failed to delete message`, ex);
+        if (!this.config.bridge.disableDeletionForwarding) {
```
Half-Shot: I would just check to see if this is true at the start of the function, and if so return early. Looks neater than wrapping the send method, and saves doing store lookups.

Half-Shot:
LGTM

PR #53, opened 2017-09-18
Set presenceInterval to min value if nil.

anoadragon453:
If presenceInterval is not present in the config, set it to the minimum allowed
value instead of having it be nil.


anoadragon453:
Also fixes #64.

Half-Shot:
LGTM 👍

PR #62, opened 2017-10-04
Fix bot starting before user/room store is initialized

Johni0702:
`bridge.run` actually returns a promise that is resolved once the user and room stores have been loaded.

Not waiting for that promise before calling `discordbot.run` can sometimes (very rarely) cause one of those stores to be accessed before it is ready, crashing the AS.


Half-Shot:
LGTM 👍

PR #71, opened 2017-10-18
HTML escape text before passing to marked

14mRh4X0r:
This prevents Matrix clients from interpreting text between angle brackets as HTML tags and thus stripping them.


Half-Shot:
Sorry 😞, I forgot this existed and it looks like marked has support for this that I missed. Closing in favor of #95.

PR #65, opened 2017-10-07
Fix puppeting (get_user_discord_ids in store.ts), fix tools/addbot.ts

eeeeeta:
This fixes the buggy code in `src/store.ts`, making `get_user_discord_ids` actually work.

While we're at it, I also changed `tools/addbot.ts` to request the necessary permissions for delete forwarding to work properly.

PR #74, opened 2017-12-18
Fix puppeting (get_user_discord_ids in store.ts), fix tools/addbot.ts

eeeeeta:
This fixes the buggy code in `src/store.ts`, making `get_user_discord_ids` actually work.

While we're at it, I also changed `tools/addbot.ts` to request the necessary permissions for delete forwarding to work properly.

*This used to be PR #65, but then I misunderstood how github works and pushed commits to the wrong branch, so you have this one now. Sorry.*

### DIFF
--- a/src/store.ts
+++ b/src/store.ts
@@ -147,9 +147,7 @@ export class DiscordStore {
       },
     ).then( (rows) => {
       if (rows !== undefined) {
-        const ret = [];
-        rows.forEach((row) => ret.push(row.discord_id));
-        return ret;
+        return rows.map((row) => row.discord_id);
       } else {
         return [];
       }
--- a/src/store.ts
+++ b/src/store.ts
@@ -147,7 +147,7 @@ export class DiscordStore {
       },
     ).then( (rows) => {
       if (rows !== undefined) {
-        let ret = [];
+        const ret = [];
         rows.forEach((row) => ret.push(row.discord_id));
         return ret;
       } else {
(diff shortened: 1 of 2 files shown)


src/store.ts
```diff
@@ -137,22 +137,24 @@ export class DiscordStore {
...
+      WHERE user_id = $userId;
       `, {
         $userId: userId,
       },
     ).then( (rows) => {
       if (rows !== undefined) {
-        rows.map((row) => row.discord_id);
+        const ret = [];
+        rows.forEach((row) => ret.push(row.discord_id));
+        return ret;
```
Half-Shot: Is there any reason why not to use map? I should be returning the result however -_-

tools/addbot.ts
```diff
@@ -21,6 +21,7 @@ const perms = flags.READ_MESSAGES |
   flags.EMBED_LINKS |
   flags.ATTACH_FILES |
   flags.READ_MESSAGE_HISTORY |
+  flags.MANAGE_MESSAGES |
```
Half-Shot: This doesn't look like it has anything to do with puppeting, so this would need to be in a seperate commit.

src/store.ts
eeeeeta: Not particularly, apart from "the bridge is broken and I know this will work, despite being stupid"...which is not really an acceptable standard for PRs, so I changed it.

tools/addbot.ts
eeeeeta: Fair enough; I've taken it out.

Half-Shot:
Looks fine to me, I should have merged this ages ago.

PR #81, opened 2018-02-13
Hs/use room member state

### LINKED ISSUE #80: Per-room display names and avatars [Matrix => Discord]
When bridging from matrix to discord, the user's m.room.member status [avatar/display name] appears to be ignored, instead of being used in order to keep the room and discord appearance of a user in sync

Half-Shot:
Fixes #80


turt2live:
image

This PR is tested in production and live on t2bot.io. Confirmed working as intended, and no unexpected consequences.

PR #85, opened 2018-02-18
Webhook description isn't allowed (apparently)

turt2live:
I'm 99% sure this isn't the correct fix, but the project doesn't build without it for me.

### DIFF
--- a/src/bot.ts
+++ b/src/bot.ts
@@ -216,7 +216,7 @@ export class DiscordBot {
       // Create a new webhook if none already exists
       try {
         if (!hook) {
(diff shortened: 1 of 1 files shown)


Half-Shot:
Not so fast? I seem to be able to build it okay. Do we have some mismatch somewhere?

turt2live:
Probably. I can't get around the error with discordjs 11.1.0 on any platform.

turt2live:
A version mismatch between our environments seems to have caused this problem.

Half-Shot:
Yeah, looks like we do require it on develop. Sorry.

PR #87, opened 2018-02-20
Package upgrades & lock file

turt2live:
I've upgraded the dependencies on t2bot.io with no issues, so this is more just housekeeping than anything.

I've also included the package-lock to help avoid problems like https://github.com/Half-Shot/matrix-appservice-discord/pull/85 in the future


package.json
```diff
@@ -31,32 +31,32 @@
...
+    "@types/bluebird": "^3.5.20",
+    "@types/node": "^7.0.55",
     "@types/sqlite3": "^2.2.32",
-    "bluebird": "^3.4.7",
+    "bluebird": "^3.5.1",
     "command-line-args": "^4.0.1",
-    "command-line-usage": "^4.0.0",
-    "discord.js": "^11.1.0",
+    "command-line-usage": "^4.1.0",
+    "discord.js": "^11.3.0",
```
turt2live: specifically, discord.js being upgraded fixes the issue described in https://github.com/Half-Shot/matrix-appservice-discord/pull/85

Half-Shot: Does it need to be bumped past 11.1.0, same for command-line-usage?

Half-Shot:
👍 for package-lock, and version bumps so long as they can be verified to be necessary.

package.json
turt2live: just discord.js being upgraded from 11.1.0 to something newer fixed my issue. This PR is more the result of me being lazy and just running `npm upgrade`.

Issue #72, opened 2017-11-17
Ability to bridge an existing matrix room to a discord channel


anoadragon453:
Perhaps by inviting the bot, and issuing the commands to bridge a discord channel to the current room.

Half-Shot:
Commenting about #90 in here. The long term plan is to split out provisioning into it's own handler. We'd have a provisioninghandler for the provisioning itself and then a webapi and commandhandler that feed into it. For the time being I am happy with #90 doing it inside matrixroomhandler and we can refactor later.

Half-Shot:
Closing this as I am happy with the existing functionality.

PR #90, opened 2018-02-22
Self-service bridging (for existing matrix rooms)

turt2live:
# Description

Addresses https://github.com/Half-Shot/matrix-appservice-discord/issues/72

This is done by adding a fairly basic command handler to the bridge. If enabled in the configuration, users around matrix may use `!discord help` in a room of their choice to get more information about how to bridge a channel to their room. By default the self-service options are disabled to prevent accidentally sharing the bot link. This is because up until now the bridge operator was the one to give out the bot link.

Users can bridge existing matrix rooms by inviting the bridge to the matrix room (and discord guild), then sending the message `!discord bridge <guild id> <channel id>`. Users need at least power level 50, in most rooms, to bridge a Discord guild. Similarly, they can terminate the bridge in the room by using `!discord unbridge`.

Only one channel is permitted per room because the bridge makes heavy use of the context API, which only provides the first `RemoteRoom` it finds. To avoid having to make the switch to a different system, I've opted for a single channel in a room (which is going to be the most popular case anyways).

Discord guild owners have a chance to approve/decline the bridge if they so choose.

### DIFF
--- a/src/matrixroomhandler.ts
+++ b/src/matrixroomhandler.ts
@@ -201,7 +201,7 @@ export class MatrixRoomHandler {
               });
 
               await this.provisioner.AskBridgePermission(channel, event.sender);
-              await this.provisioner.BridgeMatrixRoom(channel, event.room_id);
+              this.provisioner.BridgeMatrixRoom(channel, event.room_id);
               return this.bridge.getIntent().sendMessage(event.room_id, {
                   msgtype: "m.notice",
                   body: "I have bridged this room to your channel",
@@ -250,7 +250,8 @@ export class MatrixRoomHandler {
               log.error("MatrixRoomHandler", err);
               return this.bridge.getItent().sendMessage(event.room_id, {
                   msgtype: "m.notice",
-                  body: "There was an error unbridging this room. Please try again later or contact the bridge operator.",
+                  body: "There was an error unbridging this room. " +
+                    "Please try again later or contact the bridge operator.",
               });
           }
       } else if (command === "help") {
--- a/src/matrixroomhandler.ts
+++ b/src/matrixroomhandler.ts
@@ -2,6 +2,7 @@ import { DiscordBot } from "./bot";
 import {
   Bridge,
   RemoteRoom,
+  MatrixRoom,
   thirdPartyLookup,
   thirdPartyProtocolResult,
   thirdPartyUserResult,
@@ -12,19 +13,35 @@ import { DiscordBridgeConfig } from "./config";
 import * as Discord from "discord.js";
 import * as log from "npmlog";
 import * as Bluebird from "bluebird";
+import { Util } from "./util";
+import { Provisioner } from "./provisioner";
 
 const ICON_URL = "https://matrix.org/_matrix/media/r0/download/matrix.org/mlxoESwIsTbJrfXyAAogrNxA";
 const JOIN_DELAY = 6000;
 const HTTP_UNSUPPORTED = 501;
 const ROOM_NAME_PARTS = 2;
 const AGE_LIMIT = 900000; // 15 * 60 * 1000
+const PROVISIONING_DEFAULT_POWER_LEVEL = 50;
+const PROVISIONING_DEFAULT_USER_POWER_LEVEL = 0;
+
+// Note: The schedule must not have duplicate values to avoid problems in positioning.
+/* tslint:disable:no-magic-numbers */ // Disabled because it complains about the values in the array
+const JOIN_ROOM_SCHEDULE = [
+    0,              // Right away
+    1000,           // 1 second
+    30000,          // 30 seconds
+    300000,         // 5 minutes
+    900000,         // 15 minutes
+];
+/* tslint:enable:no-magic-numbers */
 
 export class MatrixRoomHandler {
+
   private config: DiscordBridgeConfig;
   private bridge: Bridge;
   private discord: DiscordBot;
   private botUserId: string;
-  constructor (discord: DiscordBot, config: DiscordBridgeConfig, botUserId: string) {
+  constructor (discord: DiscordBot, config: DiscordBridgeConfig, botUserId: string, private provisioner: Provisioner) {
     this.discord = discord;
     this.config = config;
     this.botUserId = botUserId;
@@ -74,12 +91,16 @@ export class MatrixRoomHandler {
       this.HandleInvite(event);
     } else if (event.type === "m.room.redaction" && context.rooms.remote) {
       this.discord.ProcessMatrixRedact(event);
-    } else if (event.type === "m.room.message" && context.rooms.remote) {
+    } else if (event.type === "m.room.message") {
       log.verbose("MatrixRoomHandler", "Got m.room.message event");
-      const srvChanPair = context.rooms.remote.roomId.substr("_discord".length).split("_", ROOM_NAME_PARTS);
-      return this.discord.ProcessMatrixMsgEvent(event, srvChanPair[0], srvChanPair[1]).catch((err) => {
-        log.warn("MatrixRoomHandler", "There was an error sending a matrix event", err);
-      });
+      if (event.content.body && event.content.body.startsWith("!discord")) {
+        return this.ProcessCommand(event, context);
+      } else if (context.rooms.remote) {
+        const srvChanPair = context.rooms.remote.roomId.substr("_discord".length).split("_", ROOM_NAME_PARTS);
+        return this.discord.ProcessMatrixMsgEvent(event, srvChanPair[0], srvChanPair[1]).catch((err) => {
+            log.warn("MatrixRoomHandler", "There was an error sending a matrix event", err);
+        });
+      }
     } else {
       log.verbose("MatrixRoomHandler", "Got non m.room.message event");
     }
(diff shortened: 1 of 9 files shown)


src/matrixroomhandler.ts
```diff
@@ -12,14 +13,35 @@ import { DiscordBridgeConfig } from "./config";
...
 const ICON_URL = "https://matrix.org/_matrix/media/r0/download/matrix.org/mlxoESwIsTbJrfXyAAogrNxA";
 const JOIN_DELAY = 6000;
 const HTTP_UNSUPPORTED = 501;
 const ROOM_NAME_PARTS = 2;
 const AGE_LIMIT = 900000; // 15 * 60 * 1000
+const RETRY_INVITE_ACCEPT_DELAY = 5000; // ms
 
 export class MatrixRoomHandler {
+
+  public static GetBotLink(config: any): string {
```
Half-Shot: Nit: I'm no fan of statics appearing above the properties and constructor. Could it be placed *after* instance methods?

src/matrixroomhandler.ts
```diff
@@ -74,19 +96,157 @@ export class MatrixRoomHandler {
...
   }
 
   public HandleInvite(event: any) {
-    // Do nothing yet.
+    log.info("MatrixRoomHandler", "Received invite for " + event.state_key + " in room " + event.room_id);
+    if (event.state_key === this.botUserId) {
+      this.bridge.getIntent().getClient().joinRoom(event.room_id).catch((err) => {
+        log.error("MatrixRoomHandler", err);
+        setTimeout(() => {
+            this.bridge.getIntent().join(event.room_id).catch((err2) => log.error("MatrixRoomHandler", err2));
```
Half-Shot: Need a bit more detail than just logging the error here. Especially since we are dropping to the floor after this.

src/matrixroomhandler.ts
```diff
@@ -74,19 +96,157 @@ export class MatrixRoomHandler {
...
+        }, RETRY_INVITE_ACCEPT_DELAY); // Retry the join in 5 seconds if it failed the first time
```
Half-Shot: I'm wondering if 5 seconds is enough, federated invites have a habit of going tits up.

src/matrixroomhandler.ts
```diff
@@ -74,19 +96,157 @@ export class MatrixRoomHandler {
...
       log.verbose("MatrixRoomHandler", "Got non m.room.message event");
     }
   }
 
   public HandleInvite(event: any) {
-    // Do nothing yet.
+    log.info("MatrixRoomHandler", "Received invite for " + event.state_key + " in room " + event.room_id);
+    if (event.state_key === this.botUserId) {
+      this.bridge.getIntent().getClient().joinRoom(event.room_id).catch((err) => {
+        log.error("MatrixRoomHandler", err);
```
Half-Shot: Same as below.

src/matrixroomhandler.ts
```diff
@@ -74,19 +96,157 @@ export class MatrixRoomHandler {
...
+              body: "The owner of this bridge does not permit self-service bridging.",
+          });
+      }
+
+      // Check to make sure the user has permission to do anything in the room. We can do this here
+      // because the only commands we support are self-service commands (which therefore require some
+      // level of permissions)
+      const plEvent = await this.bridge.getIntent().getClient().getStateEvent(event.room_id, "m.room.power_levels", "");
+      let userLevel = 0;
+      let requiredLevel = 50;
```
Half-Shot: Magic numbers :cry:

src/matrixroomhandler.ts
```diff
@@ -194,4 +354,16 @@ export class MatrixRoomHandler {
       remote,
     };
   }
+
+  private bridgeMatrixRoom(channel: Discord.TextChannel, roomId: string) {
+      const remote = new RemoteRoom(`discord_${channel.guild.id}_${channel.id}_bridged`);
+      remote.set("discord_type", "text");
+      remote.set("discord_guild", channel.guild.id);
+      remote.set("discord_channel", channel.id);
+      remote.set("manual_bridge", true);
```
Half-Shot: I'm not sure I like the name here. What's the official terminology for it in matrix? Plumbing? ``plumbed_room``.

src/matrixroomhandler.ts
```diff
@@ -74,19 +96,157 @@ export class MatrixRoomHandler {
...
+          }
+
+          if (!remoteRoom.data.manual_bridge) {
+              return this.bridge.getIntent().sendMessage(event.room_id, {
+                  msgtype: "m.notice",
+                  body: "This room cannot be unbridged.",
+              });
+          }
+
+          this.bridge.getRoomStore().removeEntriesByRemoteRoomId(remoteRoom.getId()).then(() => {
```
Half-Shot: Catch case where we fail to remove the entry?

src/matrixroomhandler.ts
turt2live: This is what tslint made me do :p

I can update the rule too if desired. This method also feels like it's in the wrong class anyways.

src/matrixroomhandler.ts
turt2live: Only used if the power levels event is missing or corrupt for some reason.

src/matrixroomhandler.ts
Half-Shot: Yeah, feels like a utility to me.

src/matrixroomhandler.ts
Half-Shot: Still, a little ``const PROVISIONING_DEFAULT_POWER_LEVEL`` might be nice

src/matrixroomhandler.ts
Half-Shot: FWIW, I have no idea if this CAN fail but I have been bitten by not catching things before.

turt2live:
@Half-Shot this is ready for review 🎉

anoadragon453:
Everything works from my testing. :+1:

anoadragon453:
Just a heads up that it would be nice if the bot told you why a room couldn't be bridged (i.e. if it's a plumbed room, then state it cannot be bridged because of that reason). 

Otherwise some users will be confused and we'll have to explain the difference between a normal and plumbed room over and over :P

src/provisioner.ts
```diff
@@ -0,0 +1,87 @@
...
+import { Permissions } from "discord.js";
+
+const PERMISSION_REQUEST_TIMEOUT = 300000; // 5 minutes
+
+export class Provisioner {
+
+    private bridge: Bridge;
+    private pendingRequests: { [channelId: string]: (approved: boolean) => void } = {}; // [channelId]: resolver fn
+
+    public setBridge(bridge: Bridge): void {
```
Half-Shot: Nitpick: Public functions should be capitalized.

src/matrixroomhandler.ts
```diff
@@ -74,19 +91,177 @@ export class MatrixRoomHandler {
...
+              return this.bridge.getIntent().sendMessage(event.room_id, {
+                  msgtype: "m.notice",
+                  body: "Invalid syntax. For more information try !discord help bridge",
+              });
+          }
+
+          const guildId = args[0];
+          const channelId = args[1];
+          let channel: Discord.TextChannel = null;
+          return this.discord.LookupRoom(guildId, channelId).then((result) => {
```
Half-Shot: Function is async, do we need to make this ``then``ables?

src/matrixroomhandler.ts
```diff
@@ -74,19 +91,178 @@ export class MatrixRoomHandler {
...
+              const channel = <Discord.TextChannel> discordResult.channel;
+
+              log.info("MatrixRoomHandler", `Bridging matrix room ${event.room_id} to ${guildId}/${channelId}`);
+              this.bridge.getIntent().sendMessage(event.room_id, {
+                  msgtype: "m.notice",
+                  body: "I'm asking permission from the guild administrators to make this bridge.",
+              });
+
+              await this.provisioner.AskBridgePermission(channel, event.sender);
+              await this.provisioner.BridgeMatrixRoom(channel, event.room_id);
```
Half-Shot: You don't need to await this one :)

turt2live: If we don't and bridging somehow fails then things get angry. This way it gets caught by the catch

src/matrixroomhandler.ts
```diff
@@ -74,19 +91,178 @@ export class MatrixRoomHandler {
...
+              return this.bridge.getIntent().sendMessage(event.room_id, {
+                  msgtype: "m.notice",
+                  body: "This room has been unbridged",
+              });
+          } catch (err) {
+              log.error("MatrixRoomHandler", "Error while unbridging room " + event.room_id);
+              log.error("MatrixRoomHandler", err);
+              return this.bridge.getItent().sendMessage(event.room_id, {
+                  msgtype: "m.notice",
+                  body: "There was an error unbridging this room. Please try again later or contact the bridge operator.",
```
Half-Shot: Just need to "..."+"..." this one for the line length :s

src/matrixroomhandler.ts
turt2live: oh, hmm, it doesn't return a promise.

/me removes await

Half-Shot:
Build is happy = half-shot is happy. Thanks!

PR #92, opened 2018-02-22
Standardise coding conventions in tslint.json

Half-Shot:
This PR is both a change and a discussion.

I've added a few options I would like to use and once agreed I will refactor the code to conform to that standard. This is expected to be a breaking change for a lot of peoples PRs, but necessary to avoid any more upset IDEs and to keep the code clean going forward now the project has become more busy.

### DIFF
--- a/tslint.json
+++ b/tslint.json
@@ -26,6 +26,10 @@
       "severity": "warning"
     },
     "no-unused-expression": true,
-    "prefer-const": true
+    "prefer-const": true,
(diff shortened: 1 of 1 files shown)


Sorunome:
I'd remove that max file count. I can understand the thinking behind it, but shouldn't there be stuff rather for methods having a max line count? Also, some test files would explode that, so they would need to be exempted somehow

PR #93, opened 2018-02-25
Fix checkboxes on README, check Webhooks

anoadragon453:
Noticed the checkboxes's indentation on the README for the roadmap weren't visible with GitHub's rendered markdown. Adding an indentation of 4 spaces fixes this, and makes things a lot clearer :)

I also checked webhooks, since I'm pretty sure we handle those at this point.


Half-Shot:
Will merge once we have everything else settled.

Issue #94, opened 2018-02-26
Incorrect HTML escaping


Sorunome:
formatted_body doesn't HTML escape the parts, this is what the message source looks like

This results in some clients, like riot, not rendering the `<Sorunome>` part as it should be `&lt;Sorunome&gt;`
```
  "content": {
    "body": "**<Sorunome>** test",
    "msgtype": "m.text",
    "formatted_body": "<p><strong><Sorunome></strong> test</p>\n",
    "format": "org.matrix.custom.html"
  },
```

Half-Shot:
Fixed by #95. Closing.

PR #95, opened 2018-02-26
fix html escaping, #94

Sorunome:
i tested it


Half-Shot:
Nitpick, can you put it just above the class declarations, for the sake of consistency? Thanks a bunch for the fix.

Sorunome:
Done. Also, I just noticed that this makes #71 redundant, but as this is at the library level of the marked library thing, i believe it is the cleaner solution

Half-Shot:
Indeed. This looks great. Merging.

PR #105, opened 2018-03-03
fix discord message content parsing

Sorunome:
This commit fixes that the parsed discord stuff is HTML escaped, so that e.g. emojis are displayed correctly again

### DIFF
--- a/src/messageprocessor.ts
+++ b/src/messageprocessor.ts
@@ -5,9 +5,12 @@ import { DiscordBot } from "./bot";
 import * as escapeStringRegexp from "escape-string-regexp";
 
 const USER_REGEX = /<@!?([0-9]*)>/g;
+const USER_REGEX_POSTMARK = /&lt;@!?([0-9]*)&gt;/g;
 const CHANNEL_REGEX = /<#?([0-9]*)>/g;
+const CHANNEL_REGEX_POSTMARK = /&lt;#?([0-9]*)&gt;/g;
 const EMOJI_SIZE = "1em";
 const EMOJI_REGEX = /<:\w+:?([0-9]*)>/g;
+const EMOJI_REGEX_POSTMARK = /&lt;:\w+:?([0-9]*)&gt;/g;
 const MATRIX_TO_LINK = "https://matrix.to/#/";
 
 marked.setOptions({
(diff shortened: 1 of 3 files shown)


Half-Shot:
Oh also, if you could modify some of the tests for MessageProcessor to include both postmark and non postmark? I know it's a faff but I really need to see it tested. Ideally also tests showing emojis in their markdown and non markdown format.

anoadragon453:
About to head to bed but took a quick look. Definitely fix those whitespace issues before finalizing.

Otherwise yes, tests would be good. I can test on my bridge tomorrow as well. Looking forward to this landing :)

Sorunome:
having slep over it i would actually suggest some changes: different parsing for postmark and premark. Because premark is the text equivalent (postmark as currently). Premark e.g.:
 - Channels: just `#channelname`
 - Emoji: put `:emojiname:`
 - higlights: just put the member name, instead of parsing `@_discord_${id}:blahblabhlah`


this could actually make it cleaner then, since everything has different cases, to have a `ReplaceMembers` function and a `ReplaceMembersPostmark` etc.

turt2live:
I'm not super well versed in how Discord does things, but this looks like it catches the common complaints about formatting at least.

src/messageprocessor.ts
```diff
@@ -66,46 +78,49 @@ export class MessageProcessor {
...
-              const mxcUrl = await this.bot.GetGuildEmoji(msg.guild, id);
-              content = content.replace(results[0],
-                  `<img alt="${id}" src="${mxcUrl}" style="height: ${EMOJI_SIZE};"/>`);
-          } catch (ex) {
-              log.warn("MessageProcessor",
-              `Could not insert emoji ${id} for msg ${msg.id} in guild ${msg.guild.id}: ${ex}`,
-            );
-          }
-          results = EMOJI_REGEX.exec(content);
+            const id = results[1];
```
anoadragon453: These whitespace issues, where things are being indented slightly.

Sorunome: that is because it had mixed 4-space and 2-space indentation so i put it all to 4-space

Half-Shot:
@Sorunome Looks fine to me, just need the tests to go along with it 👍

Half-Shot:
This is looking good. I'll merge 👍

PR #106, opened 2018-03-03
Broader emoji support

Sorunome:
This PR adds support for animated emoji and external emoji. For some reason only the first frame of animated emoji is been displayed (idk why x.x)

~~It builds ontop of #105~~ was merged to upstream and fixed commits in this PR

### DIFF
--- a/src/messageprocessor.ts
+++ b/src/messageprocessor.ts
@@ -10,12 +10,13 @@ const USER_REGEX_POSTMARK = /&lt;@!?([0-9]*)&gt;/g;
 const CHANNEL_REGEX = /<#?([0-9]*)>/g;
 const CHANNEL_REGEX_POSTMARK = /&lt;#?([0-9]*)&gt;/g;
 const EMOJI_SIZE = "1em";
-const EMOJI_REGEX = /<:(\w+):([0-9]*)>/g;
-const EMOJI_REGEX_POSTMARK = /&lt;:(\w+):([0-9]*)&gt;/g;
+const EMOJI_REGEX = /<(a?):(\w+):([0-9]*)>/g;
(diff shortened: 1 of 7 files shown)


src/messageprocessor.ts
```diff
@@ -66,46 +78,51 @@ export class MessageProcessor {
...
-              content = content.replace(results[0],
-                  `<img alt="${id}" src="${mxcUrl}" style="height: ${EMOJI_SIZE};"/>`);
-          } catch (ex) {
-              log.warn("MessageProcessor",
-              `Could not insert emoji ${id} for msg ${msg.id} in guild ${msg.guild.id}: ${ex}`,
-            );
-          }
-          results = EMOJI_REGEX.exec(content);
+            const animated = results[1] === "a";
+            const name = results[2];
```
Sorunome: It gives linting errors here and the following line for "no magic numbers" but i have absolutely no idea how to fix that, as, well, that is how you fetch regex match groups

Sorunome:
I figured out why animated emojis only displayed the first frame:  
As it turns out, riot creates the url for the thumbnail out of the mxc URL.

This is because in matrix-js-sdk [here](https://github.com/matrix-org/matrix-js-sdk/blob/master/src/content-repo.js#L53-L66) if any parameters are added it will fetch the thumbnail url.

Now, we aren't adding parameters ourself, but in matrix-react-sdk [here](https://github.com/matrix-org/matrix-react-sdk/blob/develop/src/HtmlUtils.js#L236-L240) when parsing the `` tag it adds defaults for width/height. I am not sure what is appropriate to fix this issue, and if that might actually be an issue with matrix-react-sdk always making mxc urls thumbnails

Half-Shot:
And coming in at nearly 2 months later, LGTM. I'll merge asap

PR #111, opened 2018-04-09
Use a thumbnailed image for sender avatars

turt2live:
Discord uses 128px square images, so 512px should be fine for most cases.


Half-Shot:
Thanks, and nice find :+1:

PR #114, opened 2018-04-12
Make sure usernames don't overflow the length requirement

turt2live:


Half-Shot:
Nice find, merging :)

PR #119, opened 2018-05-12
Match non-word characters in nicknames

### LINKED ISSUE #118: mentions don't work for nonascii characters
it is currently impossible to mention a discord user where the nickname ends or starts with a non-ascii character.

this is because javascript doesn't support unicode properly for word boundaries `\b`
as used here:
```typescript
new RegExp(
  `\\b(${matcher})(?=\\b)`
  , "mig"), `<@!${member.id}>`,
);
```
https://github.com/Half-Shot/matrix-appservice-discord/blob/develop/src/matrixeventprocessor.ts#L87

and this behavior will apparently also not change with the `/u` flag:
https://github.com/mathiasbynens/es-regexp-unicode-character-class-escapes/issues/1#issuecomment-62251775

so something like this could work
```js
/(?<=\W|^)(${matcher})(?=\W|$)/igmu
```
not using \w since that doesn't support astral characters?
but idk I'm not too familiar with typescript so you'd probably have to transpile it to something compliant with es5 and of course think around the lookbehind.

Half-Shot:
Fixes #118

### DIFF
--- a/src/matrixeventprocessor.ts
+++ b/src/matrixeventprocessor.ts
@@ -79,14 +79,14 @@ export class MatrixEventProcessor {
     }
 
     public FindMentionsInPlainBody(body: string, members: Discord.GuildMember[]): string {
+        const WORD_BOUNDARY = "(^|\:|\@|\#|```|\\s|$|,)";
         for (const member of members) {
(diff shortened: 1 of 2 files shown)


Half-Shot:
Oh um, so it looks like lookbehinds are a new feature.

Half-Shot:
Merging

PR #128, opened 2018-05-16
ClientFactory Tests

Half-Shot:


Half-Shot:
This slightly changes the order of promises and might break things, but it makes the code unracey at least.

PR #131, opened 2018-05-17
Basic support for message edits

### LINKED ISSUE #60: Bridge Message Edits
screenshot_20170930-083045

This is how the gitter bridge shows message edits without the ability to edit messages on the matrix side and I thought it was *gorgeous*. The repo for that bridge is [here](https://github.com/matrix-org/matrix-appservice-gitter).

I do not believe you can edit webhook messages, but once/if matrix gets message editing capabilities, we can send a similar message on the discord side, however perhaps with italics rather than colors, as discord is monocolor.

anoadragon453:
Needs some tests and perhaps some nicer formatting. But is functional at the moment.

image

image
--

For the sake of auditing, this fixes #60 - *Half-Shot*

### DIFF
--- a/src/bot.ts
+++ b/src/bot.ts
@@ -458,7 +458,7 @@ export class DiscordBot {
     });
 
     // Sending was a success
-    return Promise.resolve(true);
+    return true;
   }
 
   private AddGuildMember(guildMember: Discord.GuildMember) {
@@ -623,7 +623,9 @@ export class DiscordBot {
     const editedMsg = await this.msgProcessor.FormatEdit(oldMsg, newMsg);
 
     // Send the message to all bridged matrix rooms
-    this.SendMatrixMessage(editedMsg, newMsg.channel, newMsg.guild, newMsg.author, newMsg.id);
+    if (!await this.SendMatrixMessage(editedMsg, newMsg.channel, newMsg.guild, newMsg.author, newMsg.id)) {
+      log.error("DiscordBot", "Unable to announce message edit for msg id:", newMsg.id);
+    }
   }
 
     private async DeleteDiscordMessage(msg: Discord.Message) {
--- a/src/bot.ts
+++ b/src/bot.ts
@@ -588,6 +588,11 @@ export class DiscordBot {
   }
 
   private async OnUpdate(oldMsg: Discord.Message, newMsg: Discord.Message) {
+    // Check if an edit was actually made
+    if (oldMsg.toString() === newMsg.toString()) {
+      return;
+    }
+
     // Create a new edit message using the old and new message contents
     let editedMsg = await this.msgProcessor.FormatEdit(oldMsg, newMsg);
 
--- a/src/bot.ts
+++ b/src/bot.ts
@@ -80,7 +80,8 @@ export class DiscordBot {
       client.on("guildMemberAdd", (newMember) => { this.AddGuildMember(newMember); });
       client.on("guildMemberRemove", (oldMember) => { this.RemoveGuildMember(oldMember); });
       client.on("guildMemberUpdate", (_, newMember) => { this.UpdateGuildMember(newMember); });
-      client.on("messageDelete", (msg) => {this.DeleteDiscordMessage(msg); });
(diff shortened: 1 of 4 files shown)


src/bot.ts
```diff
@@ -586,6 +587,36 @@ export class DiscordBot {
     });
   }
 
+  private async OnUpdate(oldMsg: Discord.Message, newMsg: Discord.Message) {
+    // Check if an edit was actually made
+    if (oldMsg.toString() === newMsg.toString()) {
```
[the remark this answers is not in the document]
anoadragon453: Opting for `.content` instead.

anoadragon453:
This new SendMessage method is going to need a discord channel, author, messageID and guildID. All of this is contained in a `Discord.Message` object, so do we want to make a method like:

```
private SendMatrixMessageWithDiscordContext(discordMsg: Discord.Message, processorResult: MessageProcessorMatrixResult) {
```

or does that seem too messy? The method would look like:

```
  private async SendMessageWithDiscord(discordMsg: Discord.Message, matrixMsg: MessageProcessorMatrixResult) {
    const rooms = await this.GetRoomIdsFromChannel(discordMsg.channel)
    const intent = this.GetIntentFromDiscordMember(discordMsg.author);

    rooms.forEach((room) => {
      intent.sendMessage(room, {
        body: matrixMsg.body,
        msgtype: "m.text",
        formatted_body: matrixMsg.formattedBody,
        format: "org.matrix.custom.html",
      }).then((res) => {
        const evt = new DbEvent();
        evt.MatrixId = res.event_id + ";" + room;
        evt.DiscordId = discordMsg.id;
        evt.ChannelId = discordMsg.channel.id;
        evt.GuildId = discordMsg.guild.id;
        this.store.Insert(evt);
      });
    });
  }
```

anoadragon453:
However, markdown is not formatted, because if I do so, `<p>` tags end up in the front and back, messing up the `*edited*` message:

image

Any way to not have leading and ending `<p>` tags to not be inserted?

Half-Shot:
Sorry for the delay, I'm back and looking at this.

Half-Shot:
> Any way to not have leading and ending <p> tags inserted?

Your best bet really is to not do the markdown-> HTML conversion until you have concatenated both the before and after messages.

anoadragon453:
Sounds good. I'll try that and add a test for it.

src/bot.ts
```diff
@@ -434,6 +435,32 @@ export class DiscordBot {
...
+        evt.MatrixId = res.event_id + ";" + room;
+        evt.DiscordId = msgID;
+        evt.ChannelId = chan.id;
+        evt.GuildId = guild.id;
+        this.store.Insert(evt);
+      });
+    });
+
+    // Sending was a success
+    return Promise.resolve(true);
```
Half-Shot: You can probably get away with ``return true;`` since it's ``async``.

src/bot.ts
```diff
@@ -586,6 +613,19 @@ export class DiscordBot {
...
+    // Check if an edit was actually made
+    if (oldMsg.content === newMsg.content) {
+      return;
+    }
+
+    // Create a new edit message using the old and new message contents
+    const editedMsg = await this.msgProcessor.FormatEdit(oldMsg, newMsg);
+
+    // Send the message to all bridged matrix rooms
+    this.SendMatrixMessage(editedMsg, newMsg.channel, newMsg.guild, newMsg.author, newMsg.id);
```
Half-Shot: This should return

src/messageprocessor.ts
```diff
@@ -71,6 +71,12 @@ export class MessageProcessor {
         return result;
     }
 
+    public async FormatEdit(oldMsg: Discord.Message, newMsg: Discord.Message): Promise<MessageProcessorMatrixResult> {
+        // TODO: Produce a nice, colored diff between the old and new message content
+        oldMsg.content = "*edit:* ~~" + oldMsg.content + "~~ -> " + newMsg.content;
+        return await this.FormatDiscordMessage(oldMsg);
```
Half-Shot: This doesn't need to await

Half-Shot:
Okay, this all looks fine! :tada:  :taco:  :tanabata_tree:

PR #133, opened 2018-05-19
you really aren't going to like me for this [aka the 'merge eta's random crap' PR]

eeeeeta:
This merges in all the various changes, including DM support, that I've extended my fork of the bridge with over time. You won't want to merge this as it is (because it's terrible, non-compartmentalized, etc.), but it's probably worth just sending this PR in to get your opinion on some of the stuff in here.
Also the 'allow edits from maintainers' thing is on, so if you want to hijack this and make it semi-usable, that's also an option.

Changes (from the top of my head; probably non-exhaustive):

- DM support (incl. typing notifs for DM rooms)
  - also currently undocumented: you start a DM by joining `#_discord_dm_YOURUSERID_OTHERUSERSID:localpart`, where the other user has to be one of your friends
- better mentions processing
  - if the user being mentioned is bridged, ping their matrix user instead of their discord ghost
  - also does pre/postmark stuff better IIRC
- edits?
  - currently redacts the message being edited and reposts it with an `(edited)`
  - also I think the code used to say how long ago it was posted, but that's gone for some reason
- I think it handles puppeting better as a result of the DM changes?
- it bridges pins, or at least tried to once
- probably other stuff


Half-Shot:
Will be on top of this as soon as I can @eeeeeta, but I'm almost certainly going to ask these to be split into branches (if one depends on another, just tell me in the PR description). Also probably going to need some kind of tests.

Sorry, got to be careful these days not to break things :s

eeeeeta:
[closing in favour of splitting up into smaller PRs later]

PR #137, opened 2018-05-21
Warn about encrypted rooms

### LINKED ISSUE #129: Warn users about E2E
As per https://github.com/vector-im/riot-web/issues/2277, we should be rejecting encrypted rooms for plumbing. We should also be warning the user that by turning on encryption, the bridge will stop working.

Half-Shot:
Fixes #129


turt2live:
It doesn't look like this handles rooms the bridge is already in that are encrypted. Is that intentional?

Half-Shot:
@turt2live Kinda sorta, I don't know if we are allowed to post unencrypted messages to encrypted rooms which the bridge would have to do. I guess you can do it, but I thought it wasn't recommended.

OH, also we can't decrypt bot commands anyway.

PR #139, opened 2018-05-22
[WIP] Add support for custom discord emojis

anoadragon453:
This PR adds support for parsing custom discord emotes from matrix and having them show up in Discord.

image

image

WIP because I need to add a couple test cases for it, but otherwise is all working smoothly.

### DIFF
--- a/src/matrixeventprocessor.ts
+++ b/src/matrixeventprocessor.ts
@@ -11,6 +11,8 @@ import * as log from "npmlog";
 const MaxFileSize = 8000000;
 const MIN_NAME_LENGTH = 2;
 const MAX_NAME_LENGTH = 32;
+const DISCORD_EMOJI_REGEX = /:(\w+):/g;
+
 export class MatrixEventProcessorOpts {
     constructor(
         readonly config: DiscordBridgeConfig,
@@ -45,6 +47,10 @@ export class MatrixEventProcessor {
         if (this.config.bridge.disableHereMention) {
             body = body.replace(new RegExp(`@here`, "g"), "@ here");
         }
+
+        // Handle discord custom emoji
+        body = this.ReplaceDiscordEmoji(body, channel.guild);
+
         let displayName = event.sender;
         let avatarUrl = undefined;
         if (profile) {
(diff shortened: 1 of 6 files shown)


src/bot.ts
```diff
@@ -22,6 +22,7 @@ const MIN_PRESENCE_UPDATE_DELAY = 250;
 // TODO: This is bad. We should be serving the icon from the own homeserver.
 const MATRIX_ICON_URL = "https://matrix.org/_matrix/media/r0/download/matrix.org/mlxoESwIsTbJrfXyAAogrNxA";
 class ChannelLookupResult {
+  public guild: Discord.Guild;
```
Half-Shot: Channels *have* guilds so I don't really like including it in this. You can get the guild of a channel with ``channel.guild``

src/matrixeventprocessor.ts
```diff
@@ -92,6 +98,33 @@ export class MatrixEventProcessor {
...
+    public ReplaceDiscordEmoji(content: string, guild: Discord.Guild): string {
+        let results = DISCORD_EMOJI_REGEX.exec(content);
+        while (results !== null) {
+            const emojiName = results[1];
+            const emojiNameWithColons = results[0];
+
+            // Check if this emoji exists in the guild
+            if(guild.emojis[emojiName] !== null) {
+                // Replace :a: with <:a:123ID123>
+                const emojiID = this.FindDiscordEmojiIDWithName(emojiName, guild.emojis);
```
Half-Shot: I'd have just used ``guild.emojis.find((emojiCollection) => { ... });`` but I guess either is fine.

src/bot.ts
anoadragon453: You ***WISH*** you could, but nah https://discord.js.org/#/docs/main/stable/class/Channel

Unless the docs are wrong/outdated.

src/matrixeventprocessor.ts
anoadragon453: You and your JS knowledge. I'll try that, thanks :)

src/bot.ts
Half-Shot: Ah...see this is where I cheated and casted everything to a TextChannel. Given we only ever operate on TextChannels, it's a safe bet for now.

src/matrixeventprocessor.ts
```diff
@@ -92,6 +98,25 @@ export class MatrixEventProcessor {
...
+        console.log("Gonna replace")
+        let results = DISCORD_EMOJI_REGEX.exec(content);
+        while (results !== null) {
+            const emojiName = results[1];
+            const emojiNameWithColons = results[0];
+
+            // Check if this emoji exists in the guild
+            if(guild.emojis[emojiName] !== null) {
+                // Replace :a: with <:a:123ID123>
+                const emojiID = guild.emojis.find((emoji) => { return emoji.name === emojiName }).id;
```
Half-Shot: Oooh, isn't the above ``guild.emojis[emojiName]`` already an emoji object you can peel the ``id`` from. Why are you doing two lookups?

src/matrixeventprocessor.ts
```diff
@@ -92,6 +98,25 @@ export class MatrixEventProcessor {
...
 
+    public ReplaceDiscordEmoji(content: string, guild: Discord.Guild): string {
+        console.log("Gonna replace")
+        let results = DISCORD_EMOJI_REGEX.exec(content);
+        while (results !== null) {
+            const emojiName = results[1];
+            const emojiNameWithColons = results[0];
+
+            // Check if this emoji exists in the guild
+            if(guild.emojis[emojiName] !== null) {
```
Half-Shot: ``!== null`` means it will only check for null and not undefined which is probably not what you want.

PR #144, opened 2018-05-24
Clarify where to specify appservice protocols

anoadragon453:
So I just realized that `protocols: ["discord"]` is supported to go at the end of `discord-registration.yaml`, not `homeserver.yaml` aka I'm dumb.

Hopefully this makes other people not so dumb too.


Half-Shot:
Oh yes, good point. Thanks!

PR #159, opened 2018-06-05
Add a Dockerfile and supporting configuration

turt2live:
Things need to be stored in a volume, so we also add a bunch of configuration options for moving the persistence around.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,15 @@
+.git
+node_modules
+
+# This is just stuff we don't need
+.travis.yml
+.gitignore
+build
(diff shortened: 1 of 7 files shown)


Half-Shot:
This actually all looks fine to me, but I'm fairly new to docker. I'm assuming if you've tested this, then it all works and I have no complaints.

turt2live:
It's currently running in production. One thing I forgot to mention is the instructions assume you'll make a Docker hub account and set this up to be published there. Automated builds are probably the easiest, albeit slow.

PR #162, opened 2018-06-11
Thirdparty Handler

Half-Shot:
This moves some of the code from ``bot.ts`` to ThirdPartyHandler and also adds some tests.


Sorunome:
With the channel synchronizer PR #202 the question arises if this method wouldn't rather belong into the synchronizer, as it already has like `GetRoomIdsFromChannel`

Sorunome:
@Half-Shot with all the restrucutre that happened, is this still....valid?

Half-Shot:
Probably not :(

PR #164, opened 2018-06-11
Display name, topic and membership changes on Discord.

### LINKED ISSUE #155: Show kicks/bans on Discord side for Matrix rooms

Half-Shot:
Fixes #155 

This just consumes state events and sends a message from the bot into the bridged channel.

### DIFF
--- a/src/bot.ts
+++ b/src/bot.ts
@@ -182,22 +182,26 @@ export class DiscordBot {
   }
 
   public async ProcessMatrixStateEvent(event: any): Promise<void> {
-      log.verbose("DiscordBot", `Got state event from ${roomId} ${event.type}`);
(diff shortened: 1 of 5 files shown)


Sorunome:
why not attempt to set name / topic on the discord side, with fallback to what you have in there? Perhaps the bot has name / topic setting permissions after all

Sorunome:
> More settings for provisioning, like a "!track topic/name" would be :+1: 

That kinda goes into #206 

So, if i understand it correctly, what i suggested is out-of-scope of this PR?

Half-Shot:
>So, if i understand it correctly, what i suggested is out-of-scope of this PR?

For now, yes. We'd want to do this properly and that time will come when we can setup proper provisioning settings :)

PR #182, opened 2018-07-19
Replace npmlog with winston

Half-Shot:


Half-Shot:
I think this PR is ready, pending review from interested parties.

But I'm holding off until we have the major PRs merged because of the conflicts.

turt2live:
I would have settled for timestamps, but this is much more interesting :D

It would be nice to be able to disable file logging entirely and instead opt for console logging. This is in particularly useful for Docker containers which don't need the logs duplicated.

Half-Shot:
>It would be nice to be able to disable file logging entirely and instead opt for console logging. This is in particularly useful for Docker containers which don't need the logs duplicated.

It does that by default?

Half-Shot:
Yaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaymergeddddd

PR #142, opened 2018-05-23
Hs/updateuser.improvements

Half-Shot:
This change removes the UpdateUser function and instead spreads out that functionality to a UserSync class where we make the algorithm more verbose and easier to test. 

What does this do:

* Users are now joined and left from the guilds appropriately
* Profile updates are made, and guild nicks overwrite them automatically.
* Pulled out more code from ``bot.ts``

Fixes #56, #101 and #102

PR #185, opened 2018-07-19
changing username respects guild overrides

Sorunome:
When a user on the discord side changes the username now, the nicknames are only in the bridged rooms updated where there isn't a separate guild nickname override

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,3 +45,4 @@ discord-registration.yaml
 build
 
 *.db
+*.db.backup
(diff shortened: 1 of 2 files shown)


Sorunome:
Any idea why the build failed? I developed in node 10.6.0

Half-Shot:
This was somewhat obsoleted by #142. Admittedly the PR is a blob, but it's part of our approach to move logic out of ``bot.ts`` and also to have more testable verbose logic.

But thank you, you found bug #188 which is going to be a **lot of fun** :(

PR #186, opened 2018-07-19
fix message processing

Sorunome:
Fixes a bunch of message processing things, e.g. discord doesn't have quotes, tables, headers, urls etc. and those were still passed through marked

### DIFF
--- a/src/matrixeventprocessor.ts
+++ b/src/matrixeventprocessor.ts
@@ -58,6 +58,9 @@ export class MatrixEventProcessor {
             body = `*${body}*`;
         }
 
+        // replace <del>blah</del> with ~~blah~~
+        body = body.replace(/<del>([^<]*)<\/del>/g, "~~$1~~");
+
         // Handle discord custom emoji
         body = this.ReplaceDiscordEmoji(body, channel.guild);
 
--- a/src/messageprocessor.ts
+++ b/src/messageprocessor.ts
@@ -17,23 +17,30 @@ const ANIMATED_EMOJI_REGEX_GROUP = 1;
 const NAME_EMOJI_REGEX_GROUP = 2;
 const ID_EMOJI_REGEX_GROUP = 3;
 
-marked.setOptions({
-    sanitize: true,
-    tables: false,
-});
-
-const markedLexer = new marked.Lexer();
-for (const r of ["hr", "heading", "lheading", "blockquote", "list", "item", "bullet", "def", "table", "lheading"]) {
-    markedLexer.rules[r] = /$^/;
+function _setupMarked() {
+    marked.setOptions({
+        sanitize: true,
+        tables: false,
+    });
+    
+    const markedLexer = new marked.Lexer();
+    // as discord doesn't support these markdown rules
+    // we want to disable them by setting their regexes to non-matchable ones
+    // deleting the regexes would lead to marked-internal errors
+    for (const r of ["hr", "heading", "lheading", "blockquote", "list", "item", "bullet", "def", "table", "lheading"]) {
+        markedLexer.rules[r] = /$^/;
+    }
+    // paragraph-end matching is different, as we don't have headers and thelike
+    markedLexer.rules.paragraph = /^((?:[^\n]+\n\n)+)\n*/;
+    
+    const markedInlineLexer = new marked.InlineLexer(true);
+    // same again, remove tags discord doesn't support
+    for (const r of ["tag", "link", "reflink", "nolink", "br"]) {
+        markedInlineLexer.rules[r] = /$^/;
+    }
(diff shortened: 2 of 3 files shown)


Half-Shot:
Looks mostly good, what does it do with tables? Removes the content or just the tags?

Sorunome:
```
 marked.setOptions({
     sanitize: true,
     tables: false,
 });
```

As discord markdown doesn't have tables, it is disabled, along with quotes, headings and so much more.

So, if someone on discord sends
```
1 | 2
-----
3 | 4
```
it will appear as just that on the matrix side

src/matrixeventprocessor.ts
```diff
@@ -58,6 +58,9 @@ export class MatrixEventProcessor {
             body = `*${body}*`;
         }
 
+        // replace <del>blah</del> with ~~blah~~
+        body = body.replace(/<del>([^<]*)<\/del>/g, "~~$1~~");
```
Half-Shot: Simple, I like :+1:

src/messageprocessor.ts
```diff
@@ -19,8 +19,22 @@ const ID_EMOJI_REGEX_GROUP = 3;
 
 marked.setOptions({
     sanitize: true,
+    tables: false,
 });
 
+const markedLexer = new marked.Lexer();
+for (const r of ["hr", "heading", "lheading", "blockquote", "list", "item", "bullet", "def", "table", "lheading"]) {
+    markedLexer.rules[r] = /$^/;
```
Half-Shot: Could you leave a comment here to say what this is doing?

src/messageprocessor.ts
```diff
@@ -19,8 +19,22 @@ const ID_EMOJI_REGEX_GROUP = 3;
...
+}
+markedLexer.rules.paragraph = /^((?:[^\n]+\n\n)+)\n*/;
+
+const markedInlineLexer = new marked.InlineLexer(true);
+for (const r of ["tag", "link", "reflink", "nolink", "br"]) {
+    markedInlineLexer.rules[r] = /$^/;
+}
+markedInlineLexer.rules.em = /^_([^_](?:[^_]|__)*?[^_]?)_\b|^\*((?:\*\*|[^*])+?)\*(?!\*)/;
```
Half-Shot: Nit pick about 20-35 -> I'd prefer this to be in a ``function _setupMarked()`` and then called at the bottom of the file, only because I feel like it's a bit messy.

src/messageprocessor.ts
```diff
@@ -47,20 +61,19 @@ export class MessageProcessor {
...
-        // for both plaintext and markdown
-        content = this.InsertEmbeds(content, msg);
         
         // for the formatted body we need to parse markdown first
         // as else it'll HTML escape the result of the discord syntax
-        let contentPostmark = marked(content);
+        let contentPostmark = marked(content).replace(/\n/g, "<br>").replace(/(<br>)?<\/p>(<br>)?/g, "</p>");
         
         // parse the plain text stuff
+        content = this.InsertEmbeds(content, msg);
```
Half-Shot: I see that this got added but I don't see any tests for InsertEmbed?

Sorunome:
embeds never had tests. I could add some with this PR, if you want

Am 22. Juli 2018 11:43:49 MESZ schrieb Will Hunt <notifications@github.com>:

src/messageprocessor.ts
Half-Shot: Why did we move InsertEmbeds? The comment above suggests we really don't want to be doing that because otherwise we are inserting embeds after generating the HTML. (Though I must admit the tests didn't break, which is concerning)

Sorunome:
embeds used header and URL markdown, which discord markdown doesn't have and thus is disabled

Am 22. Juli 2018 11:45:48 MESZ schrieb Will Hunt <notifications@github.com>:

Half-Shot:
>embeds never had tests. I could add some with this PR, if you want

That's kinda bad of me! At any rate, we need tests for them if we are changing behaviours!

Half-Shot:
>embeds used header and URL markdown, which discord markdown doesn't have and thus is disabled

But the messageprocessor outputs messages for Matrix, which does support them.

Sorunome:
>> embeds used header and URL markdown, which discord markdown doesn't have and thus is disabled

> But the messageprocessor outputs messages for Matrix, which does support them.

While that is correct, `marked()` here simulates the discord markdown flavour, thus passing the embed content through `marked()` wouldn't get the URLs and `<h5>` parsed, thus i splitted it into normal and postmark like with the other methods

Half-Shot:
Hmm, okay! Let's just ensure the tests cover the behaviour.

PR #189, opened 2018-07-22
User fixes

### LINKED ISSUE #188: On first use of an Intent, it will join the room which will render previous roomstate broken.
(This is branched off #56)

There are several ways we can fix this:
1. Fix synapse/the spec not to nuke memberstate by somehow pinning important keys.
2. Fix matrix-appservice-bridge to not autojoin the room if it doesn't know the state of the room.
3. Fix the above to store memberstate in a file rather than in memory.
4. Don't use matrix-appservice-bridge intents, roll our own.

Sorunome:
Fixes #188 and ensures users are in a room when sending a message

### DIFF
--- a/src/bot.ts
+++ b/src/bot.ts
@@ -524,19 +524,27 @@ export class DiscordBot {
       if (msg.content !== null && msg.content !== "") {
         this.msgProcessor.FormatDiscordMessage(msg).then((result) => {
             rooms.forEach((room) => {
-              intent.sendMessage(room, {
+              const trySend = () => intent.sendMessage(room, {
                 body: result.body,
                 msgtype: "m.text",
                 formatted_body: result.formattedBody,
                 format: "org.matrix.custom.html",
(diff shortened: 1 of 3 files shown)


Half-Shot:
Works for me, just wondering if we do any other room actions which might need the same treatment, or if we need to register the user..

Oh and tests would be handy

Sorunome:
Technically things like typing indicator, message deleting and editing need the same treatment, practically i thought that after sending a message stuff should be fine.

As for tests....i don't really know how to write tests for this :[

Half-Shot:
I'm wondering if we should be rolling our own intent handler rather than trying to hack around the bridge sdk...hmm.

Sorunome:
I personally think rolling an own intent handler would be more work.

Half-Shot:
Aha, but travis has made it more enticing https://github.com/turt2live/matrix-js-bot-sdk

PR #190, opened 2018-07-22
update presence handler

Sorunome:
Updates presence handler to register the user on M_FORBIDDEN instead of constantly giving out warnings


Half-Shot:
I'd be a little surprised that we don't register everyone on creating a new channel now, since we join everyone. *However* this is probably fine. Do we have a test to ensure the bot isn't put in the presence queue :innocent:

Sorunome:
Good point, I'll add a test :+1:

PR #175, opened 2018-07-09
Add custom keys for discord member

Half-Shot:
Another request from /dev/ponies, this is adding a few custom keys from discord into the member object.

Fixes #173

PR #194, opened 2018-07-23
Soru/usernamefix

Sorunome:
This makes sure that, when a new bridged room is created, the nickname is set correctly on the ghosts instead of username#hashtag being used.

It also provides a script to fix existing rooms

### DIFF
--- a/src/matrixroomhandler.ts
+++ b/src/matrixroomhandler.ts
@@ -87,7 +87,11 @@ export class MatrixRoomHandler {
             return this.discord.UserSyncroniser.OnUpdateUser(member.user);
         }).then(() => {
             log.info("OnAliasQueried", `Joining ${member.id} to ${roomId}`);
-            return this.joinRoom(this.discord.GetIntentFromDiscordMember(member), roomId);
+            return this.joinRoom(this.discord.GetIntentFromDiscordMember(member), roomId)
+                .then(() => {
+                  // set the correct discord guild name
+                  this.discord.UserSyncroniser.EnsureJoin(member, roomId);
+                });
         }));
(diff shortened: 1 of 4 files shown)


src/matrixroomhandler.ts
```diff
@@ -87,7 +87,11 @@ export class MatrixRoomHandler {
             return this.discord.UserSyncroniser.OnUpdateUser(member.user);
         }).then(() => {
             log.info("OnAliasQueried", `Joining ${member.id} to ${roomId}`);
-            return this.joinRoom(this.discord.GetIntentFromDiscordMember(member), roomId);
+            return this.joinRoom(this.discord.GetIntentFromDiscordMember(member), roomId)
+                .then(() => {
+                  // set the correct discord guild name
+                  this.discord.UserSyncroniser.EnsureJoin(member, roomId);
```
Half-Shot: This looks a bit bizarre. We are joining and then calling ``EnsureJoin``?

Sorunome: That is correct, as the normal join doesn't set the display name of the guild correctly. EnsureJoin will also ensure that the member has the correct nickname override for that guild set and, wtih #175 also the role stuff

Half-Shot:
Looks good, again just a curious function call. Also, does this address any issues on GH?

src/matrixroomhandler.ts
Half-Shot: Fair nuff, we probably should rethink that name but let's do that later.

Sorunome:
> Also, does this address any issues on GH?

I don't think an issue on github was created on this.

PR #195, opened 2018-07-23
use correct name when mentioning

Sorunome:
Problem: When mentioning a user on discord who has a nickname set the created pill on the riot side contained their username, not their nickname

This problem wasn't visible in desktop riot due to desktop riot ignoring what the actual contents were, however, it is very visible on android riot

### DIFF
--- a/src/messageprocessor.ts
+++ b/src/messageprocessor.ts
@@ -141,7 +141,7 @@ export class MessageProcessor {
             const id = results[1];
             const member = msg.guild.members.get(id);
             const memberId = `@_discord_${id}:${this.opts.domain}`;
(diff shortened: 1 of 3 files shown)


Half-Shot:
Works for me. Do we test this?

Sorunome:
What do you mean? It fixes the issue i outlined in the top post, you can notice the issue if you hit "View Source" on the matrix message.

riot-web does some extra parsing on user pills so riot-desktop didn't show that bug, but riot-android did. And possibly other clients.

Sorunome:
oh, you meant automated test thingies, added that now

Sorunome:
I guess me just derped, hang on

Issue #205, opened 2018-07-26
Cannot run admin tools (getbotlink, adminme) from Docker image


Pneumaticat:
The Docker image doesn't have package.json, which is necessary for npm, nor does it have any of the JavaScript files from the tools/ directory, so you can't run `npm run-script getbotlink` or `npm run-script adminme` as described in the docs.

Half-Shot:
@turt2live ?

adrianmalacoda:
You should be able to run the tools directly through `node`, i.e. as `node ./build/tools/adminme.js`. The Dockerfile should perhaps put these tools on the PATH though to make them more convenient to use (e.g. `docker exec $container_name adminme`.

So far as I can tell the Docker image contains `build` which is where the *executable* form of the tools resides (`tools` is a TypeScript source code directory)

turt2live:
This seems to be one of the cases that I simply forgot about when contributing the docker image in the first place. Personally I use different tooling to do the tasks that the scripts provide, and forgot that they existed in the project :(

although, I do seem to have forced the container to ignore the folder for some reason: https://github.com/Half-Shot/matrix-appservice-discord/blob/develop/.dockerignore#L15

Half-Shot:
#208 has been merged, closing.

PR #208, opened 2018-07-29
Assorted Dockerfile fixes and improvements

adrianmalacoda:
See #205 

* Remove `tools` directory from `.dockerignore` so it's included in the image
* Add links to all tools on the `PATH`, such that they can be executed with `docker exec`
* Set `WORKDIR` to `/data` so the tools will work correctly
* Correct path of the `discordas.js` in the `CMD`

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -12,4 +12,3 @@ README.md
 test
 docs
 config.yaml
-tools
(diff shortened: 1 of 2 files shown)


Half-Shot:
This looks like it does the right thing, happy to merge.

(@Sorunome do you want a looky or nah?)

Sorunome:
while i haven't used docker yet ever.......from my very limited knowledge it looks like it does the right things, as in, the config stuff seems to make sense, heh

Half-Shot:
We put our faith in you @adrianmalacoda

PR #210, opened 2018-07-29
Discord-side moderation of matrix users

Sorunome:
This PR allows admins of a discord guild to kick, ban and unban members on the matrix side.

`!matrix help` output:

```
Available Messages:
 - kick <name>: Kicks a user on the matrix side
 - ban <name>: Bans a user on the matrix side
 - unban <name>: Unbans a user on the matrix side

The name must be the display name or the mxid of the user to perform the action on.
```

### DIFF
--- a/src/matrixroomhandler.ts
+++ b/src/matrixroomhandler.ts
@@ -386,7 +386,6 @@ export class MatrixRoomHandler {
     return Promise.reject({err: "Unsupported", code: HTTP_UNSUPPORTED});
   }
 
-  
   public async HandleDiscordCommand(msg: Discord.Message) {
     if (!msg.member.hasPermission("ADMINISTRATOR")) {
       msg.channel.sendMessage("**ERROR:** insufficiant permissions to use matrix commands");
--- a/src/matrixroomhandler.ts
+++ b/src/matrixroomhandler.ts
@@ -378,7 +378,7 @@ export class MatrixRoomHandler {
 
   public async HandleDiscordCommand(msg: Discord.Message) {
     if (!(<Discord.TextChannel> msg.channel).guild) {
-      msg.channel.sendMessage("**ERROR:** only available for guild channels");
+      msg.channel.send("**ERROR:** only available for guild channels");
     }
     
     const {command, args} = Util.MsgToArgs(msg.content, "!matrix");
@@ -445,7 +445,7 @@ export class MatrixRoomHandler {
     }
     
     if (!msg.member.hasPermission(actions[command].permission as any)) {
-      msg.channel.sendMessage("**ERROR:** insufficiant permissions to use this matrix command");
+      msg.channel.send("**ERROR:** insufficiant permissions to use this matrix command");
       return;
     }
     
--- a/src/matrixroomhandler.ts
+++ b/src/matrixroomhandler.ts
@@ -386,6 +386,140 @@ export class MatrixRoomHandler {
     return Promise.reject({err: "Unsupported", code: HTTP_UNSUPPORTED});
   }
 
+  
+  public async HandleDiscordCommand(msg: Discord.Message) {
+    if (!msg.member.hasPermission("ADMINISTRATOR")) {
+      msg.channel.sendMessage("**ERROR:** insufficiant permissions to use matrix commands");
+      return;
+    }
+    if (!(<Discord.TextChannel> msg.channel).guild) {
+      msg.channel.sendMessage("**ERROR:** only available for guild channels");
+    }
+    const prefix = "!matrix ";
+    let command = "help";
+    let args = [];
+    if (msg.content.length >= prefix.length) {
+      const allArgs = msg.content.substring(prefix.length).split(" ");
+      if (allArgs.length && allArgs[0] !== "") {
+        command = allArgs[0];
+        allArgs.splice(0, 1);
+        args = allArgs;
+      }
+    }
+    
+    let replyMessage = "Error, unkown command. Try `!matrix help` to see all commands";
+    
+    const intent = this.bridge.getIntent();
+    const doAction = async (funcKey, action) => {
+      const name = args.join(" ");
+      const channels = await this.discord.GetRoomIdsFromChannel(msg.channel);
+      try {
+        const userMxid = await this.GetMxidFromName(name, channels);
+        let allChannels = [];
+        /* tslint:disable:await-promise */
+        await Bluebird.all((<Discord.TextChannel> msg.channel).guild.channels.map((c) => {
+        /* tslint:enable:await-promise */
+          return this.discord.GetRoomIdsFromChannel(c).then((chans) => {
+            allChannels = allChannels.concat(chans);
+          }).catch((e) => {
+            // pass, non-text-channel
+          });
+        }));
+        let errorMsg = "";
+        /* tslint:disable:await-promise */
+        await Bluebird.all(allChannels.map((c) => {
+        /* tslint:enable:await-promise */
+          return intent[funcKey](c, userMxid).catch((e) => {
+            // maybe we don't have permission to kick/ban/unban...?
+            errorMsg += `\nCouldn't ${funcKey} ${userMxid} from ${c}`;
+          });
+        }));
+        if (errorMsg) {
+          throw Error(errorMsg);
+        }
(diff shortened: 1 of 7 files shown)


Sorunome:
tslint is at it again, saying that `Bluebird.all()` wouldn't be a promise function so we can't `await` it......any suggestions, @Half-Shot ?

Half-Shot:
I just stick an ignore on it. Though Promise.all is a thing it will accept.

src/matrixroomhandler.ts
```diff
@@ -386,6 +386,139 @@ export class MatrixRoomHandler {
     return Promise.reject({err: "Unsupported", code: HTTP_UNSUPPORTED});
   }
 
+  public async HandleDiscordCommand(msg: Discord.Message) {
+    if (!msg.member.hasPermission("ADMINISTRATOR")) {
```
Half-Shot: This requirement seems a bit extreme, I'd check permissions *per command* since we might have commands that regulars might want.

src/matrixroomhandler.ts
```diff
@@ -386,6 +386,139 @@ export class MatrixRoomHandler {
...
+    if (msg.content.length >= prefix.length) {
+      const allArgs = msg.content.substring(prefix.length).split(" ");
+      if (allArgs.length && allArgs[0] !== "") {
+        command = allArgs[0];
+        allArgs.splice(0, 1);
+        args = allArgs;
+      }
+    }
+    
+    let replyMessage = "Error, unkown command. Try `!matrix help` to see all commands";
```
Half-Shot: Typo :)

src/matrixroomhandler.ts
```diff
@@ -386,6 +386,139 @@ export class MatrixRoomHandler {
...
+    
+    const intent = this.bridge.getIntent();
+    const doAction = async (funcKey, action) => {
```
Half-Shot: I wonder if this would be better as a private function rather than embedded in a function, would make for prettier typescript

Sorunome: I don't think it'd be too good to make that a private, as it is pretty specifically tailored.While it takes use of `intent` and `replyMessage` could be a return value, i have the feeling that for it being a private it should have some more abstract output maybe, not something such specific....i'm not sure

src/matrixroomhandler.ts
```diff
@@ -386,6 +386,139 @@ export class MatrixRoomHandler {
...
+
+  private async GetMxidFromName(name: string, channels: string[]) {
+    if (name[0] === "@" && name.includes(":")) {
+        return name;
+    }
+    const client = this.bridge.getIntent().getClient();
+    const matrixUsers = {};
+    let matches = 0;
+    /* tslint:disable:await-promise */
+    await Bluebird.all(channels.map((c) => {
```
Half-Shot: You can just use ``Promise.all`` here?

src/matrixroomhandler.ts
```diff
@@ -386,6 +386,139 @@ export class MatrixRoomHandler {
...
+        await doAction("unban", "Unbanned");
+        break;
+      default:
+        break;
+    }
+    
+    msg.channel.send(replyMessage);
+  }
+
+  private async GetMxidFromName(name: string, channels: string[]) {
```
Half-Shot: This *definitely* feels like it belongs in Util

Half-Shot: Is this a list of discord channels or roomids, the variable name isn't helpful.

src/matrixroomhandler.ts
Half-Shot: I have a passionate hate for variables that don't describe what they are. ``c`` should be at least ``chan`` or ``channel``. I suspect it's a roomId though.

src/matrixroomhandler.ts
```diff
@@ -386,6 +386,139 @@ export class MatrixRoomHandler {
...
+    const matrixUsers = {};
+    let matches = 0;
+    /* tslint:disable:await-promise */
+    await Bluebird.all(channels.map((c) => {
+    /* tslint:enable:await-promise */
+      // we would use this.bridge.getBot().getJoinedMembers()
+      // but we also want to be able to search through banned members
+      // so we gotta roll our own thing
+      return client._http.authedRequestWithPrefix(
+        undefined, "GET", "/rooms/" + encodeURIComponent(c) + "/members",
```
Half-Shot: All these undefined's make me sad, but I don't know what we can do about it :/

We should at least have each of these on a newline though

src/matrixroomhandler.ts
```diff
@@ -386,6 +386,139 @@ export class MatrixRoomHandler {
...
+        undefined, undefined, "/_matrix/client/r0",
+      ).then((res) => {
+        res.chunk.forEach((member) => {
```
Half-Shot: What happens if this is chunked and the user is not found in the first chunk?

src/matrixroomhandler.ts
```diff
@@ -386,6 +386,139 @@ export class MatrixRoomHandler {
...
+      try {
+        const userMxid = await this.GetMxidFromName(name, channels);
+        let allChannels = [];
+        /* tslint:disable:await-promise */
+        await Bluebird.all((<Discord.TextChannel> msg.channel).guild.channels.map((c) => {
+        /* tslint:enable:await-promise */
+          return this.discord.GetRoomIdsFromChannel(c).then((chans) => {
+            allChannels = allChannels.concat(chans);
+          }).catch((e) => {
+            // pass, non-text-channel
```
Half-Shot: Do we want to at least log this?

src/matrixroomhandler.ts
```diff
@@ -386,6 +386,139 @@ export class MatrixRoomHandler {
...
+    let replyMessage = "Error, unkown command. Try `!matrix help` to see all commands";
+    
+    const intent = this.bridge.getIntent();
+    const doAction = async (funcKey, action) => {
+      const name = args.join(" ");
+      const channels = await this.discord.GetRoomIdsFromChannel(msg.channel);
+      try {
+        const userMxid = await this.GetMxidFromName(name, channels);
+        let allChannels = [];
+        /* tslint:disable:await-promise */
```
Half-Shot: [You can certainly lint rules for one line](https://github.com/Half-Shot/matrix-appservice-discord/blob/master/src/discordas.ts#L24), btw.

src/matrixroomhandler.ts
Sorunome: What do you mean? There can be multiple chunks / we need multiple `/members` calls under some circumstances?

Half-Shot:
Wow this is looking much slicker

Half-Shot:
Just need tests for bans and unbans?

test/test_util.ts
```diff
@@ -0,0 +1,119 @@
...
+                "/rooms/abc/members": [
+                    {
+                        membership: "join",
+                        mxid: "@123:localhost",
+                        displayname: "GoodBoy",
+                    },
+                ],
+            };
+            const intent = CreateMockIntent(mockRooms);
+            expect(Util.GetMxidFromName(intent, "badboy", ["abc"])).to.eventually.be.rejected;
```
Half-Shot: I think ``.eventually.`` calls need to return

src/util.ts
```diff
@@ -132,6 +152,91 @@ export class Util {
...
+        return name;
+    }
+    const client = intent.getClient();
+    const matrixUsers = {};
+    let matches = 0;
+    await Promise.all(channelMxids.map((chan) => {
+      // we would use this.bridge.getBot().getJoinedMembers()
+      // but we also want to be able to search through banned members
+      // so we gotta roll our own thing
+      return client._http.authedRequestWithPrefix(
```
Half-Shot: Eek, this looks expensive :(

Sorunome: It doesn't get called frequently, so i thought it wasn't too big a concern? Any suggestion ideas, though? (we did discuss exactly this in the matrix room once o.o)

Half-Shot:
I might have said "sod it, let's see how it goes" with TravisR weeping in the corner.

Sorunome:
> I might have said "sod it, let's see how it goes" with TravisR weeping in the corner.

Yep, iirc it was that way

Issue #201, opened 2018-07-24
Current highlight regex can cause unwanted mentions on Discord


Teteros:
Problem: Matrix style mentions have no prefix, other than the colon `:` character being appended on tab completion in some clients.

It's possible to unknowingly mention any discord member which username/nick matches one of the whole words sent to discord.

e.g: If I type:
> You can change your **nick** with /nick Nickname.

on matrix, and someone called **Nick** is a member on the Discord server, they will be mentioned:
`You can change your @nick with /nick Nickname`

This would also be the case if someone named say... **change** or **you** and so on existed of course.
To avoid this you need to escape known words with `code` tags.

Unfortunately this issue can happen quite often on popular discord servers, especially ones with no email verification on as anyone can pick a short or random common english word as their name.

Possible solutions that come to mind:
Using discord style `@mentions` on matrix (not convenient)
Relying on tab completion style mentions: e.g `nick:`
Excluding short words from being able to mention or checking if someone is a member before pinging?

Sorunome:
i personally really like the idea with doing a discord mention only if you tab-complete the pill. This, however, has two issues:
1st: you can't make pills on mobile
2nd: it's hard to implement as pills are only visible in `formatted_body` and we use `body` to parse the message

the 1st can be circumvented by highlighting on the discord side both on pills and with @nickname

Half-Shot:
FYI, we had the opposite problem in #46

Sorunome:
I think the idea of this issue is to only highlight when you tab complete or long-hold a users avatar on mobile to get their name, not when manually typing it out, as the later results in a lot of accidental highlights.

Teteros:
This popped up again in [discussions](https://matrix.to/#/!OluOzqbTJgecdDIuKn:half-shot.uk/$1541725081170424IOKFS:swedneck.xyz) today and I'll note second issue linked to this one:

> Q: Say you have (users) `word#1111` and `word#2222` on discord
and you type `word` on matrix, who gets pinged (on discord)?
A: Currently whoever is first in the userlist array (the internal one)

In the long term it might be better to use pills (parse mxid from `formatted_body`) to get the exact user to mention rather than use bare display name.

What's holding this back for now is client support (riot-android and others doesn't do pills, only `mention:` in `body`)

Sorunome:
merged

PR #211, opened 2018-07-30
Simple tab completion mention feature

adrianmalacoda:
See #201 

This is a simple tab-completion mention implementation that I cooked up for my instance. As i'm not really sure what the ideal implementation of mentions should be, I made a separate setting for toggling this particular implementation, with the idea of perhaps adding Discord-style at-mentions as a separate thing.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -17,7 +17,7 @@ export class DiscordBridgeConfig {
       if (typeof(configLayer[key]) === "object")  {
         this.ApplyConfig(newConfig[key], this[key]);
         return;
-      } 
+      }
       configLayer[key] = newConfig[key];
     });
   }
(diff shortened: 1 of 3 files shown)


Sorunome:
I'm a bit confused as to how this solves that issue, as it doesn't remove the user replacement of other plaintext stuff.

Also, I'm not sure we were done discussing that....while I'm personally in favour of having stricter highlight rules, i'm not sure what others stances are on that

adrianmalacoda:
It doesn't touch the existing mentions replacement because the discussion on that wasn't done, so I wasn't sure what the preferred solution would be. This was just something I cooked up to scratch my own itch, as I ran into that same problem early on. This is probably a sign that I jumped the gun in submitting this, so I'll go ahead and close it for now.

Issue #212, opened 2018-07-30
Docker image fails on startup: ./config/config.schema.yaml not found


Pneumaticat:
Since I recently pulled from develop, I now get this error:

```
Unable to load crypto module: crypto will be disabled: Error: global.Olm is not defined
Loading config file /data/config.yaml
Init Failed to start bridge.
Init { Error: ENOENT: no such file or directory, open './config/config.schema.yaml'
    at Object.openSync (fs.js:443:3)
    at Object.readFileSync (fs.js:348:35)
    at ConfigValidator._loadFromFile (/node_modules/matrix-appservice-bridge/lib/components/config-validator.js:54:29)
    at new ConfigValidator (/node_modules/matrix-appservice-bridge/lib/components/config-validator.js:15:28)
    at Cli._loadConfig (/node_modules/matrix-appservice-bridge/lib/components/cli.js:158:21)
    at Cli._assignConfigFile (/node_modules/matrix-appservice-bridge/lib/components/cli.js:144:23)
    at Cli.run (/node_modules/matrix-appservice-bridge/lib/components/cli.js:138:10)
    at Object.<anonymous> (/build/src/discordas.js:22:9)
    at Module._compile (internal/modules/cjs/loader.js:689:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:700:10)
    at Module.load (internal/modules/cjs/loader.js:599:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:538:12)
    at Function.Module._load (internal/modules/cjs/loader.js:530:3)
    at Function.Module.runMain (internal/modules/cjs/loader.js:742:12)
    at startup (internal/bootstrap/node.js:266:19)
    at bootstrapNodeJSCore (internal/bootstrap/node.js:596:3)
  errno: -2,
  syscall: 'open',
  code: 'ENOENT',
  path: './config/config.schema.yaml' }
```

It looks like https://github.com/Half-Shot/matrix-appservice-discord/blob/3db3d4732f9fd4fe36bc82bc62884b08f49c0d59/src/discordas.ts#L14 checks for `./config/config.schema.yaml`, but changes in #208 to set the `WORKDIR` to /data mean that it now looks in `/data/config/config.schema.yaml`, which obviously doesn't exist.

I'm not sure of a good way to fix this without breaking the tools in Docker again by reverting the WORKDIR change.

adrianmalacoda:
[...]

PR #214, opened 2018-07-30
CD to /data before running tool, so that discord-registration.yaml is found

adrianmalacoda:
See #205 #212 

The tools do need to be run from `/data` so they can find the `discord-registration.yaml`.


Sorunome:
Thanks again for the quick fix!

PR #217, opened 2018-07-31
Logging/Config testing

Half-Shot:
and a minor bugfix to config

### DIFF
--- a/test/test_log.ts
+++ b/test/test_log.ts
@@ -1,0 +1,77 @@
+import * as Chai from "chai";
+import * as ChaiAsPromised from "chai-as-promised";
+import * as Proxyquire from "proxyquire";
+import * as RealLog from "../src/log";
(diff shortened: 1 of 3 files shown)


test/test_log.ts
```diff
@@ -0,0 +1,77 @@
...
+describe("Log", () => {
+
+    beforeEach(() => {
+        loggerClosed = false;
+        loggedMessages = [];
+    });
+
+    describe("ConfigureBridge", () => {
+        it("should pass if config is empty", () => {
+            Log.ConfigureBridge({});
```
Sorunome: This has no checks, shouldn't you do something like try...catch to expect it to pass?

Half-Shot: It should fail if it throws. We can't really test the config because the config is just the default which might change.

PR #220, opened 2018-08-01
have puppet do deletions, if present

Sorunome:
Problem: If the matrix link bot on the discord side lacks the "manage messages" permissions, it is impossible to relay redacts over

While the PR doesn't solve that completely, if you have a puppeted account, it will delete the message via the puppet, and as you are allowed to delete your own messages that will still work

### DIFF
--- a/src/bot.ts
+++ b/src/bot.ts
@@ -257,25 +257,19 @@ export class DiscordBot {
     }
     log.info(`Got redact request for ${event.redacts}`);
     log.verbose(`Event:`, event);
+    
     const storeEvent = await this.store.Get(DbEvent, {matrix_id: event.redacts + ";" + event.room_id});
+    
     if (!storeEvent.Result) {
       log.warn(`Could not redact because the event was not in the store.`);
       return;
     }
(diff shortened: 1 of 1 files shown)


src/bot.ts
```diff
@@ -257,25 +257,20 @@ export class DiscordBot {
...
-        return;
-      }
-      if (!this.bot.guilds.get(storeEvent.GuildId).channels.has(storeEvent.ChannelId)) {
-        log.warn(`Could not redact because the guild could not be found.`);
-        return;
-      }
-      const channel = <Discord.TextChannel> this.bot.guilds.get(storeEvent.GuildId)
-                      .channels.get(storeEvent.ChannelId);
-      const msg = await channel.fetchMessage(storeEvent.DiscordId);
+      const result = await this.LookupRoom(storeEvent.GuildId, storeEvent.ChannelId, event.sender);
```
Half-Shot: What happens is this fails, should we report it?

Sorunome: That already catches and falls back to the normal bot within that function, so i don't think it is needed

PR #221, opened 2018-08-01
Adds support for stickers

Sorunome:
Adds support to send stickers over to discord

### DIFF
--- a/src/matrixeventprocessor.ts
+++ b/src/matrixeventprocessor.ts
@@ -40,6 +40,10 @@ export class MatrixEventProcessor {
                 channel.members.array(),
             );
 
+        if (event.type === "m.sticker") {
+            body = "";
+        }
+
         // Replace @everyone
         if (this.config.bridge.disableEveryoneMention) {
             body = body.replace(new RegExp(`@everyone`, "g"), "@ everyone");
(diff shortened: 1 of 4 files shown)


src/matrixroomhandler.ts
```diff
@@ -117,9 +117,9 @@ export class MatrixRoomHandler {
       return this.discord.ProcessMatrixStateEvent(event);
     } else if (event.type === "m.room.redaction" && context.rooms.remote) {
       return this.discord.ProcessMatrixRedact(event);
-    } else if (event.type === "m.room.message") {
-        log.verbose("Got m.room.message event");
-        if (event.content.body && event.content.body.startsWith("!discord")) {
+    } else if (event.type === "m.room.message" || event.type === "m.sticker") {
+        log.verbose("Got m.room.message / m.sticker event");
```
Half-Shot: That's just laaaazy, use ``Got ${event.type} event``

Sorunome: ooops my brain somehow turned off

src/matrixeventprocessor.ts
```diff
@@ -166,7 +170,9 @@ export class MatrixEventProcessor {
             "m.video",
             "m.file",
             "m.sticker",
-        ].indexOf(event.content.msgtype) !== -1;
+        ].indexOf(event.content.msgtype) !== -1 || [
```
Half-Shot: We have ``includes``! `indexOf() == -1 is old school` :)

PR #244, opened 2018-09-24
Start /me command with user display name

### LINKED ISSUE #112: /me is inconsistent
if John uses `/me`, the resulting message on matrix will look like this:
`*John slaps Jim around a bit with a large trout `
while on discord it looks like this:
`slaps Jim around a bit with a large trout `

anoadragon453:
Fixes #112 

It breaks the flow a little bit for users to see a /me command without a username in front. I believe the reason for the initial implementation was so that we followed Discord's implementation of the /me command, but from asking users it seems that having it go in one way from Matrix and coming out a different way in Discord can make messages sound different and thus confusion in conversation.

This PR aligns both platforms to see the same content.


src/matrixeventprocessor.ts
```diff
@@ -99,7 +99,14 @@ export class MatrixEventProcessor {
...
 
         if (event.content.msgtype === "m.emote") {
-            body = `*${body}*`;
+          if (profile &&
+              profile.displayname &&
+              profile.displayname.length >= MIN_NAME_LENGTH &&
+              profile.displayname.length <= MAX_NAME_LENGTH) {
+            body = `* \* ${profile.displayname} ${body}*`;
+          } else {
+            body = `* ${body} *`;
```
Half-Shot: What's with the spacing?

anoadragon453:
Ah, did this all on my phone termux/vim due to no laptop. Spacing looked good in the editor :P

And do we really need tests for this?

Half-Shot:
>And do we really need tests for this?

Always, otherwise we risk breaking it by accident.

>Spacing looked good in the editor :P 
``body = `* ${body} *`;``

PR #246, opened 2018-10-11
Use membership cache for EnsureJoin

Half-Shot:
For a bit of background, the membership cache is an exposed cache between the bridge lib and the bridge itself which allows us to peek in at who we think is already joined by what state events we've seen and what joins/leaves we've sent. 

This will also mean we don't set state in some cases which might change the per room custom profiles.


Sorunome:
Maybe use this PR to also rename `EnsureJoin` to something more accurate?

Half-Shot:
#308 makes this kinda obsolete

PR #250, opened 2018-10-22
Allow files missing `info`

Half-Shot:


Half-Shot:
>looks like a reasonable thing

\- @turt2live

Issue #245, opened 2018-10-05
Drop npmlog


Half-Shot:
A relic from times past, we don't use it anymore

PR #251, opened 2018-10-23
Multibacked support for databases

### LINKED ISSUE #158: Support PostgreSQL as an option
The bridge I run is approaching capacity at 22k+ active users due to the bridge not being able to acquire locks on the database fast enough. At the very least, switching to postgres would alleviate some of the concern, even without optimizing the code for it.

Half-Shot:
This is a PR to attempt to escape `node-sqlite3`, which is no longer building reliably. This also starts work on making generic connectors rather than a single database instance by creating a connector class.

I've gotten the tests to pass but can't be sure this is working 100% yet, testers welcome!

**EDIT:** Due to feature creeping the hell out of this, it now supports postgres too! I know right!

Also I got fed up of ``npmlog`` and removed it from all the leftover places.
 
Fixes #158, #245.


turt2live:
Migration steps would be nice too :)

PR #258, opened 2018-10-24
Ignore commands if the bot isn't joined to the room.

### LINKED ISSUE #257: Don't answer bot commands not addressed to the bot's userid
Bots from across the fedisphere join if they hear a "!discord" via the event stream because a user from the same homeserver happens to be in the room.

Half-Shot:
Fixes #257

### DIFF
--- a/src/matrixroomhandler.ts
+++ b/src/matrixroomhandler.ts
@@ -120,7 +120,10 @@ export class MatrixRoomHandler {
       return this.discord.ProcessMatrixRedact(event);
     } else if (event.type === "m.room.message" || event.type === "m.sticker") {
         log.verbose(`Got ${event.type} event`);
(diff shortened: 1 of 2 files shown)


src/matrixroomhandler.ts
```diff
@@ -172,6 +171,15 @@ export class MatrixRoomHandler {
   }
 
   public async ProcessCommand(event: any, context: any) {
+      const intent = this.bridge.getIntent();
+      // Due to #257 we need to check if we are joined.
+      try {
+          await intent.getClient().sendReadReceipt(event.event_id);
```
turt2live: why not check the joinedRooms?

Half-Shot: Partly because I think doing a sendReciept would be a bit quicker than trying to pull all the joined rooms from synapse and keeping a cache up to date, and partly because now the bot gives feedback when it's read the message albeit fairly pointlessly.

Issue #238, opened 2018-09-05
Usernames not parsed in Message Replies


anoadragon453:
![image](https://user-images.githubusercontent.com/1342360/45096866-5f1c9a00-b0d6-11e8-9201-3a332dc14d7b.png)

Usernames are not parsed into Discord compatible usernames in replies properly. Here is another example.

![del1](https://user-images.githubusercontent.com/1342360/45096950-925f2900-b0d6-11e8-8c3c-2c8e0d448cd9.png)
![del1](https://user-images.githubusercontent.com/1342360/45096964-9e4aeb00-b0d6-11e8-8e78-2b90d22d7b2c.png)

Not sure if it's due to the `>` in front or the bot not parsing correctly.

Sorunome:
It is because that is exactly how the markdown for the replies looks like. I agree that it needs better displaying on the discord side

Sorunome:
done in  #259

Issue #187, opened 2018-07-19
Bridge replies as embeds


Half-Shot:
We should actually switch to HTML parsing first and handle the format:

```
<mx-reply><blockquote><a href="matrix.to/event">In reply to</a> <a href="matrix.to/author">Original Author</a><br><p>Message content</p>\n</blockquote></mx-reply><p>message reply content</p>\n
```

Sorunome:
done in  #259

PR #259, opened 2018-10-25
Show replies as embeds

Half-Shot:


Half-Shot:
Fixes #238  #187

src/matrixeventprocessor.ts
```diff
@@ -14,6 +14,7 @@ const MaxFileSize = 8000000;
 const MIN_NAME_LENGTH = 2;
 const MAX_NAME_LENGTH = 32;
 const DISCORD_EMOJI_REGEX = /:(\w+):/g;
+const REPLY_REGEX = /> <(@.*:.*)> (.*)\n\n(.*)/;
```
Sorunome: This regex could be better, matching MXIDs rather than anything. Also it appears that the reply content is wrapped into `""`

src/matrixeventprocessor.ts
```diff
@@ -212,6 +206,78 @@ export class MatrixEventProcessor {
...
+        // Try to get the event.
+        try {
+            const sourceEvent = await intent.getEvent(eventId);
+            let replyText = sourceEvent.content.body  || "Reply with unknown content";
+            // Check if this is also a reply.
+            if (sourceEvent.content && sourceEvent.content["m.relates_to"] &&
+                sourceEvent.content["m.relates_to"]["m.in_reply_to"]) {
+                const sourceMatch = REPLY_REGEX.exec(sourceEvent.content.body);
+                if (sourceMatch && sourceMatch.length === REPLY_MATCHES) {
+                    replyText = sourceMatch[INDEX_REPLY];
```
Sorunome: message 1: <stuff>
message 2: <reply to message 1>
message 3: <reply to message 2>

we are replying to message 2, though, and not message 1, so why bother to get that source?

src/matrixeventprocessor.ts
```diff
@@ -212,6 +206,78 @@ export class MatrixEventProcessor {
...
+                }
+            }
+            embed.setDescription(replyText);
+            this.SetEmbedAuthor(
+                embed,
+                sourceEvent.sender,
+                await intent.getProfileInfo(sourceEvent.sender),
```
Sorunome: idea: short-circuit `@_discord_uiserid:homeserver` MXIDs and somehow link their discord profile

PR #261, opened 2018-10-25
add chanfix

Sorunome:
Adds a chanfix script which will **force** update of channel name and topic. **optionally**, if needed, it updates the icon


Half-Shot:
LGTM

PR #292, opened 2018-11-07
Soru/mocha

Sorunome:
Updates mocha

```
sorunome@sorunome-desktop repos/matrix-appservice-discord $ npm audit
                                                                                
                       === npm audit security report ===                        
                                                                                
found 0 vulnerabilities
 in 1182 scanned packages
```

adds `why-is-node-running` because newer mocha-version doesn't auto-quit, and then you can easily test for floating promises

async/await'ifies all tests

fix code & tests where floating promises existed

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ logs
 *.log.*
 npm-debug.log*
 .audit.json
+*-audit.json
 
 # Runtime data
 pids
--- a/test/config.ts
+++ b/test/config.ts
@@ -1,5 +1,8 @@
-import {argv} from "process";
-import {Log} from "../src/log";
+import { argv } from "process";
+import { Log } from "../src/log";
+import * as WhyRunning from "why-is-node-running";
+
+const logger = new Log("MessageProcessor");
 
 // we are a test file and thus need those
 /* tslint:disable:no-unused-expression max-file-line-count */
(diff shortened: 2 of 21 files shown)


Half-Shot:
Do we need chai-as-promised if all the tests are async now?

Sorunome:
> Do we need chai-as-promised if all the tests are async now?

It turns out soru couldn't find out how to async-ify the last test of the user synchroniser, so that currently still uses chai-as-promised :/

test/config.ts
```diff
@@ -1,9 +1,16 @@
...
+const logger = new Log("MessageProcessor");
 
 // we are a test file and thus need those
 /* tslint:disable:no-unused-expression max-file-line-count */
 
 if (!argv.includes("--noisy")) {
     Log.ForceSilent();
 }
+
+after(() => {
```
Half-Shot: Wait `after(()=>)` waits for the tests to all finish and then run things, that's pretty cool :o. Would this fire every time? I'm wondering if we should have some accompanying logging.

.gitignore
```diff
@@ -6,6 +6,7 @@ logs
 *.log.*
 npm-debug.log*
 .audit.json
+*-audit.json
```
Half-Shot: huh?

test/config.ts
Sorunome: it fires once after all tests are done!

.gitignore
Sorunome: it appears npm audit log filename patterns changed with some version of npm

Half-Shot: oh lovely

PR #295, opened 2018-11-08
Fix message queuing so it won't fall over if a promise rejects

Half-Shot:


src/bot.ts
```diff
@@ -112,23 +112,29 @@ export class DiscordBot {
           await Bluebird.delay(this.config.limits.discordSendDelay);
           this.discordMessageQueue[msg.channel.id] = (async () => {
               await (this.discordMessageQueue[msg.channel.id] || Promise.resolve());
-              await this.DeleteDiscordMessage(msg);
+              await (this.DeleteDiscordMessage(msg).catch(
```
Sorunome: please use

```
try {
    await blubb();
} catch (err) {
    // stuffs
}
```
instead....and basically everywhere

src/bot.ts
```diff
@@ -112,23 +112,29 @@ export class DiscordBot {
...
+              ));
           })();
       });
       client.on("messageUpdate", async (oldMessage: Discord.Message, newMessage: Discord.Message) => {
           // tslint:disable-next-line:await-promise
           await Bluebird.delay(this.config.limits.discordSendDelay);
           this.discordMessageQueue[newMessage.channel.id] = (async () => {
               await (this.discordMessageQueue[newMessage.channel.id] || Promise.resolve());
-              await this.OnMessageUpdate(oldMessage, newMessage);
+              await (this.OnMessageUpdate(oldMessage, newMessage).catch(
```
Sorunome: here too

src/bot.ts
```diff
@@ -112,23 +112,29 @@ export class DiscordBot {
...
+              ));
           })();
       });
       client.on("message", async (msg: Discord.Message) => {
         // tslint:disable-next-line:await-promise
         await Bluebird.delay(this.config.limits.discordSendDelay);
         this.discordMessageQueue[msg.channel.id] = (async () => {
             await (this.discordMessageQueue[msg.channel.id] || Promise.resolve());
-            await this.OnMessage(msg);
+            await (this.OnMessage(msg).catch(
```
Sorunome: here too

Issue #173, opened 2018-06-24
Add extra information keys to m.room.member events for members


Half-Shot:
/dev/ponies has nefarious use cases for this, such as colouring users in their fork of Riot. We should support putting in discord keys into the body such as: 

``com.discordapp.member.rolecolour``
``com.discordapp.member.bot``

anoadragon453:
![](http://static.comicvine.com/uploads/original/14/144443/4509795-3444595_640px.jpg)

I am on the case.

Half-Shot:
I fear for the project

Sorunome:
got implemented

Issue #227, opened 2018-08-06
More custom keys on m.room.member


Sorunome:
add keys so that it's like:
```
"uk.half-shot.discord.member": {
    id: <user id>,
    username: "user#1234",
    is_bot: true|false,
    roles: [],
},
```

Sorunome:
got implemented

PR #303, opened 2018-11-16
add more custom member keys

Sorunome:
implements #173


Sorunome:
and also implements #227

PR #307, opened 2018-11-16
ignore embeds with same url as in body

Sorunome:
This prevents discords auto-generated embed from URL previews to be transmitted to matrix, resulting in multiple matrix-sided url previews

### DIFF
--- a/src/bot.ts
+++ b/src/bot.ts
@@ -581,7 +581,7 @@ export class DiscordBot {
                     await this.store.Insert(evt);
                 });
             });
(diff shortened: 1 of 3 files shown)


Sorunome:
only if the embed links the same url.

we could also, in addition, check if the author is a user, and not a bot

Half-Shot:
Oh naaah, that's probably sane. LGTM. Also this fixes one of our bugs?

Half-Shot:
Oh, apparently not. Anyway, cool.

PR #308, opened 2018-11-16
make clear what EnsureJoin is about and re-write it to make it clear

### LINKED ISSUE #230: EnsureJoin is too expensive
It will send a state event each time it is called, which will usually no-op but that's quite unacceptable at scale.

Sorunome:
Removes `EnsureJoin` and splits it up to `JoinRoom` and `SetRoomState`.

`JoinRoom` is meant to, well, join a user to a room, on a basic level (it handles inviting, if needed, and also sets all the custom `m.member` keys, which is why matrixroomhadnlers `joinRoom` function (the one with the join scheduler) uses this now)

`SetRoomState` is meant for updating the custom `m.member` keys, e.g. on role changes (not implemented yet, thus currently unused).

This should show that #246 kinda is unneeded, and should also clarify / close #230

### DIFF
--- a/src/usersyncroniser.ts
+++ b/src/usersyncroniser.ts
@@ -140,9 +140,29 @@ export class UserSyncroniser {
         }
     }
 
-    public async EnsureJoin(member: GuildMember, roomId: string) {
+    public async JoinRoom(member: GuildMember, roomId: string) {
         const state = await this.GetUserStateForGuildMember(member, "");
-        log.info(`Ensuring ${state.id} is joined to ${roomId}`);
+        log.info(`Joining ${state.id} in ${roomId}`);
+        try {
+            await this.ApplyStateToRoom(state, roomId, member.guild.id);
+        } catch (e) {
+            if (e.errcode !== "M_FORBIDDEN") {
+                log.warn(`Failed to join ${state.id} to ${roomId}`, e);
+            } else {
+                log.info(`User not in room ${roomId}, inviting`);
+                try {
+                    await this.bridge.getIntent().invite(roomId, state.mxUserId);
+                    await this.ApplyStateToRoom(state, roomId, member.guild.id);
+                } catch (e) {
+                    log.error(`Failed to join ${state.id} to ${roomId}`, e);
+                }
+            }
+        }
+    }
+
+    public async SetRoomState(member: GuildMember, roomId: string) {
+        const state = await this.GetUserStateForGuildMember(member, "");
+        log.info(`Setting room state for ${state.id} in ${roomId}`);
         await this.ApplyStateToRoom(state, roomId, member.guild.id);
     }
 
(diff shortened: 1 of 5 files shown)


src/usersyncroniser.ts
```diff
@@ -140,9 +140,31 @@ export class UserSyncroniser {
...
+                log.error(`Failed to join ${state.id} to ${roomId}`, e);
+                throw e;
+            } else {
+                log.info(`User not in room ${roomId}, inviting`);
+                try {
+                    await this.bridge.getIntent().invite(roomId, state.mxUserId);
+                    await this.ApplyStateToRoom(state, roomId, member.guild.id);
+                } catch (e) {
+                    log.error(`Failed to join ${state.id} to ${roomId}`, e);
+                    throw e;
```
Half-Shot: Does this not get caught in the surrounding catch one level below, and therefore JoinRoom will not throw (which ideally it should)

src/usersyncroniser.ts
```diff
@@ -338,7 +345,8 @@ export class UserSyncroniser {
             return UserSyncroniser.ERR_CHANNEL_MEMBER_NOT_FOUND;
         }
         const state = await this.GetUserStateForGuildMember(member, ev.content!.displayname);
-        return this.ApplyStateToRoom(state, roomId, member.guild.id);
+        await this.ApplyStateToRoom(state, roomId, member.guild.id);
+        return "";
```
Half-Shot: wha? why do we not want to return the result of ApplyStateToRoom? This function must return a IUserState

test/test_matrixroomhandler.ts
```diff
@@ -87,7 +87,7 @@ function createRH(opts: any = {}) {
         },
     };
     const us = {
-        EnsureJoin: async () => { },
+        JoinRoom: async () => { USERSJOINED++; },
```
Half-Shot: Are we testing this?

src/usersyncroniser.ts
Sorunome: we need it to throw so that the schedule joiner in matrixRoomHandler can do its job

test/test_matrixroomhandler.ts
Sorunome: Yes, implicitly in `joinRoom` of `matrixroomhandler`

src/usersyncroniser.ts
Sorunome: actually it needs to return a string

Sorunome: it's odd, because sorus changes here are unrelated to that but tsc started to say "nope can't do this" so soru fixed it here

`ApplyStateToRoom` returns void

should soru add `public static readonly ERR_NO_ERROR = "";` and return that?

Half-Shot: Sounds fine, yeah

src/usersyncroniser.ts
Half-Shot: Sure but if the user fails to join, then this function won't throw and that's bad. You have a catch inside a catch.

Sorunome: It should still throw, as it is in the catch block, not the try block.

```
try {
   throw "blubb"; // falls back to catch
} catch (e) {
    throw e; // throws an outermost error
}
```

and (labeling try-1 and try-2 incorrectly for easier meaning stuffs)

```
try-1 {
    throw "blubb"; // throws error, landing in catch-1
} catch-1 (e) {
    try-2 {
        throw "blubb"; // throws an error, landing in catch-2
    } catch-2 (e) {
        throw e; // this is outside of any try.....so throw it all the way
    }
}
```

Half-Shot: Yes sorry I'm having a senior moment. That makes sense :)

Half-Shot:
This appears sane, LGTM :)

PR #311, opened 2018-11-17
don't send blank messages

Sorunome:
don't send messages that have no embeds and no content at all


Half-Shot:
A noble goal. Merged

Issue #296, opened 2018-11-13
get rid of all unhandled promise exceptions


Sorunome:
use try....catch with displaying useful stack trace etc. on the outer levels (discord event handling, etc)

Sorunome:
got merged

PR #313, opened 2018-11-17
add try...catch everywhere

Sorunome:
to remove unhandled promise rejections, solving #296

### DIFF
--- a/src/bot.ts
+++ b/src/bot.ts
@@ -106,41 +106,41 @@ export class DiscordBot {
             client.on("typingStart", async (c, u) => {
                 try {
                     await this.OnTyping(c, u, true);
-                } catch (err) { log.error("typingStart", err); }
+                } catch (err) { log.warning("Exception thrown while handling \"typingStart\" event", err); }
             });
             client.on("typingStop", async (c, u) => {
                 try {
                     await this.OnTyping(c, u, false);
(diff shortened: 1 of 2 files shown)


src/bot.ts
```diff
@@ -103,64 +103,117 @@ export class DiscordBot {
         const client = await this.clientFactory.getClient();
 
         if (!this.config.bridge.disableTypingNotifications) {
-            client.on("typingStart", async (c, u) => this.OnTyping(c, u, true) );
-            client.on("typingStop", async (c, u) => this.OnTyping(c, u, false) );
+            client.on("typingStart", async (c, u) => {
+                try {
+                    await this.OnTyping(c, u, true);
+                } catch (err) { log.error("typingStart", err); }
```
Half-Shot: Some of these could be warnings, because someone on error logging probably doesn't care if the odd typing notif fails.

src/bot.ts
```diff
@@ -103,64 +103,117 @@ export class DiscordBot {
...
+            });
+            client.on("typingStop", async (c, u) => {
+                try {
+                    await this.OnTyping(c, u, false);
+                } catch (err) { log.error("typingStop", err); }
```
Half-Shot: These errors could stand to be a little more descriptive like "Exception thrown while handling "typingStop" event"

Issue #301, opened 2018-11-14
Send Matrix HTML as Discord Markdown


swedneck:

Half-Shot:
This is a pretty undescriptive issue

Sorunome:
merged

Issue #163, opened 2018-06-11
Filter out rogue HTML from messages


Half-Shot:
@tulir has created pretty rainbox text using ``<font>`` and in the process made discord look super ugly.

Sorunome:
I think that the issue is that, in the normal `body` attribute that html is, not only in `formatted_body` when sending stuff.

Half-Shot:
Yes, that certainly doesn't help.

Sorunome:
merged

Issue #82, opened 2018-02-13
Escaping matrix event content causes problems if the text IS plain


Half-Shot:
23adab0d9a980849d848cc0ff253f4dead2453eb introduced a regression where if the text isn't markdownable, then discord assumes the ``\`` at the start of the string is intended to be part of the message.

So a string like "\*test*" will be correctly shown as "*test*" but "\test" will show up as "\test"

Basically, Discord sucks and doesn't have a good way of escaping markdown. I'm going revert this behaviour until we can think of a good workround.

Sorunome:
merged

Issue #315, opened 2018-11-18
Convert Matrix's @\room to @\everyone or @\here in Discord


Sorunome:
Convert @ room to @everyone or @here when sending? Discuss!

TildeSlashC0re:
voting for @here

Half-Shot:
I concur with @TildeSlashC0re

Sorunome:
merged

PR #316, opened 2018-11-18
HTML to markdown parsing for M->D

Sorunome:
this introduces an HTML to markdown parser for sending messages M->D

in the background it uses `node-html-parser` ( https://www.npmjs.com/package/node-html-parser )

This tackles #301, #201, #163 and #82, and should maintenance and bugfixes in the future easier.

### DIFF
--- a/src/bot.ts
+++ b/src/bot.ts
@@ -5,7 +5,11 @@ import { DbEmoji } from "./db/dbdataemoji";
 import { DbEvent } from "./db/dbdataevent";
 import { MatrixUser, RemoteUser, Bridge, Entry, Intent } from "matrix-appservice-bridge";
 import { Util } from "./util";
-import { MessageProcessor, MessageProcessorOpts, MessageProcessorMatrixResult } from "./messageprocessor";
+import {
+    DiscordMessageProcessor,
+    DiscordMessageProcessorOpts,
+    DiscordMessageProcessorMatrixResult,
+} from "./discordmessageprocessor";
 import { MatrixEventProcessor, MatrixEventProcessorOpts } from "./matrixeventprocessor";
 import { PresenceHandler } from "./presencehandler";
 import { Provisioner } from "./provisioner";
@@ -49,7 +53,7 @@ export class DiscordBot {
     private bridge: Bridge;
     private presenceInterval: number;
     private sentMessages: string[];
-    private msgProcessor: MessageProcessor;
+    private discordMsgProcessor: DiscordMessageProcessor;
     private mxEventProcessor: MatrixEventProcessor;
     private presenceHandler: PresenceHandler;
     private userSync: UserSyncroniser;
@@ -64,8 +68,8 @@ export class DiscordBot {
         this.store = store;
         this.sentMessages = [];
         this.clientFactory = new DiscordClientFactory(store, config.auth);
-        this.msgProcessor = new MessageProcessor(
-            new MessageProcessorOpts(this.config.bridge.domain, this),
+        this.discordMsgProcessor = new DiscordMessageProcessor(
+            new DiscordMessageProcessorOpts(this.config.bridge.domain, this),
         );
         this.presenceHandler = new PresenceHandler(this);
         this.discordMessageQueue = {};
@@ -446,7 +450,7 @@ export class DiscordBot {
         return rooms.map((room) => room.matrix.getId());
     }
 
-    private async SendMatrixMessage(matrixMsg: MessageProcessorMatrixResult, chan: Discord.Channel,
+    private async SendMatrixMessage(matrixMsg: DiscordMessageProcessorMatrixResult, chan: Discord.Channel,
                                     guild: Discord.Guild, author: Discord.User,
                                     msgID: string): Promise<boolean> {
         const rooms = await this.channelSync.GetRoomIdsFromChannel(chan);
@@ -584,7 +588,7 @@ export class DiscordBot {
             if (msg.content === null) {
                 return;
             }
-            const result = await this.msgProcessor.FormatDiscordMessage(msg);
+            const result = await this.discordMsgProcessor.FormatDiscordMessage(msg);
             if (!result.body) {
                 return;
             }
@@ -629,7 +633,7 @@ export class DiscordBot {
         }
 
         // Create a new edit message using the old and new message contents
-        const editedMsg = await this.msgProcessor.FormatEdit(oldMsg, newMsg);
+        const editedMsg = await this.discordMsgProcessor.FormatEdit(oldMsg, newMsg);
 
         // Send the message to all bridged matrix rooms
         if (!await this.SendMatrixMessage(editedMsg, newMsg.channel, newMsg.guild, newMsg.author, newMsg.id)) {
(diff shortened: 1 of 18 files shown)


Sorunome:
now also handles #274

Sorunome:
It should be noted that this also changes emoji bridging behaviour: it no longer scrapes `:emote:` from the plain text.

It now checks the `mxc` source of an `` tag if that is present in the store and found in the guild, it uses that. Next it uses `alt` or `title` from the tag and checks if those match.

Sorunome:
Now also tackles #315

tulir:
List and blockquote handling seems to be missing. They can be nested, so fully featured handling might be a bit complex.

Some notes that might be useful:
* For blockquotes, you probably just need to prepend `> ` to each line.
* For sorted lists, you need to prepend the index followed by a dot (or `)`) and a space to the first line of each item, then indent the remaining lines. My formatters use `length(largest_index) + n` spaces. Remember to account for the ordered list start attribute (`largest_index = ol.start + length(ol.children)`).
* For unsorted lists, you need to prepend a bullet followed by a space to the first line of each item, then indent the remaining lines (`n` spaces). If you do nesting, you might want to cycle between bullets, e.g. `["●", "○", "■", "‣"]` or just use `*` for all levels
* `n` spaces is 2 for monospace and ~4 for not-monospace.

Sorunome:
Done!

Please note that the failed tests are in the D->M parser, as the external git library changed since then, and those tests are fixed in #317

Half-Shot:
FYI, tests are still failing :(

Sorunome:
> FYI, tests are still failing :(

Yes, because this was forked off of develop branch from *before* #317 was merged

src/bot.ts
```diff
@@ -5,7 +5,11 @@ import { DbEmoji } from "./db/dbdataemoji";
 import { DbEvent } from "./db/dbdataevent";
 import { MatrixUser, RemoteUser, Bridge, Entry, Intent } from "matrix-appservice-bridge";
 import { Util } from "./util";
-import { MessageProcessor, MessageProcessorOpts, MessageProcessorMatrixResult } from "./messageprocessor";
+import {
+    DiscordMessageProcessor,
+    DiscordMessageProcessorOpts,
+    DiscordMessageProcessorMatrixResult,
```
Half-Shot: I wonder if this could have just been named `DiscordMessageProcessorResult` to save some characters.

src/bot.ts
```diff
@@ -272,15 +276,8 @@ export class DiscordBot {
         const result = await this.LookupRoom(guildId, channelId, event.sender);
         const chan = result.channel;
         const botUser = result.botUser;
-        let profile = null;
```
Half-Shot: We no longer need the profile for anything?

src/bot.ts
```diff
@@ -435,6 +432,14 @@ export class DiscordBot {
         return dbEmoji.MxcUrl;
     }
 
+    public async GetEmojiByMxc(mxc: string): Promise<DbEmoji> {
```
Half-Shot: I wonder if this should have gone on `Store`

Half-Shot: Or better yet just implemented as a private function for whoever needed it.

src/matrixeventprocessor.ts
```diff
@@ -85,93 +92,29 @@ export class MatrixEventProcessor {
     }
 
     public async EventToEmbed(
-        event: IMatrixEvent, profile: IMatrixEvent|null, channel: Discord.TextChannel,
+        event: IMatrixEvent, channel: Discord.TextChannel, getReply: boolean = true,
     ): Promise<IMatrixEventProcessorResult> {
-        let body: string = this.config.bridge.disableDiscordMentions ? event.content!.body as string :
```
Half-Shot: Are we no longer mention matching?

src/matrixeventprocessor.ts
```diff
@@ -85,93 +92,29 @@ export class MatrixEventProcessor {
...
-        if (event.type === "m.sticker") {
-            body = "";
-        }
-
-        // Replace @everyone
-        if (this.config.bridge.disableEveryoneMention) {
-            body = body.replace(new RegExp(`@everyone`, "g"), "@ everyone");
+        const mxClient = this.bridge.getClientFactory().getClientAs();
+        const profile = await mxClient.getStateEvent(event.room_id, "m.room.member", event.sender);
+        if (!profile) {
```
Half-Shot: if we fail, might still be nice to get the profile of the user.

src/matrixmessageprocessor.ts
```diff
@@ -0,0 +1,281 @@
...
+        const CHANNEL_REGEX = /^#_discord_[0-9]*_([0-9]*)/;
+        const match = id.match(CHANNEL_REGEX);
+        if (!match || !this.guild.channels.get(match[1])) {
+            return MATRIX_TO_LINK + id;
+        }
+        return `<#${match[1]}>`;
+    }
+
+    private async parsePillContent(node: Parser.HTMLElement): Promise<string> {
+        const attrs = node.attributes;
```
Half-Shot: Worth mentioning that this breaks non-pilling clients and bridges.

src/matrixmessageprocessor.ts
```diff
@@ -0,0 +1,281 @@
...
+    }
+
+    private escapeDiscord(msg: string): string {
+        if (this.opts.disableEveryone) {
+            msg = msg.replace(/@everyone/g, "@ everyone");
+        }
+        if (this.opts.disableHere) {
+            msg = msg.replace(/@here/g, "@ here");
+        }
+        msg = msg.replace(/@room/g, "@here");
```
Half-Shot: Question: This still potentially allows anyone to spam here on the discord side?

src/bot.ts
Sorunome: That moved into matrixeventprocessor.ts `EventToEmbed` which was the only place where it was used. That was also an elegant fix for something with replies

src/bot.ts
Sorunome: Soru put it there as the normal `GetEmoji` method was already there, so if one is moved, both should be - should they be moved? Maybe this would belong better into a different PR?

src/matrixeventprocessor.ts
Sorunome: Nope, we had discussed this in the matrix room, to only use pills for user mentions.
screenshot_2018-11-30_11-02-29

src/matrixmessageprocessor.ts
Sorunome: see comment above on mentions

src/matrixmessageprocessor.ts
Sorunome: yes, this way everyone who says `@room` would get that replaced to `@here`. Soru thought the proper permission parsing stuffs would fit better in a different PR

src/bot.ts
Half-Shot: Eh, leave it for now and we can come back to it when we can.

src/matrixmessageprocessor.ts
Sorunome: it's in #328

PR #317, opened 2018-11-18
convert @everyone and @here

### LINKED ISSUE #314: D->M @everyone @here
transform those to @ room

Sorunome:
D -> M

* converts @everyone and @here to @room
* update `discord-markdown`
* properly parse plaintext embeds

Question: by default @room only highlights if you are at least level 50 (moderator), what do?

fixes #314

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "bluebird": "^3.5.1",
     "command-line-args": "^4.0.1",
     "command-line-usage": "^4.1.0",
-    "discord-markdown": "git+https://github.com/Sorunome/discord-markdown.git#ca9d27039d977b35aba39cd9173e69a428118d3f",
+    "discord-markdown": "^2.0.0",
     "discord.js": "^11.4.2",
     "escape-html": "^1.0.3",
     "escape-string-regexp": "^1.0.5",
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "bluebird": "^3.5.1",
     "command-line-args": "^4.0.1",
     "command-line-usage": "^4.1.0",
-    "discord-markdown": "git+https://github.com/Sorunome/discord-markdown.git",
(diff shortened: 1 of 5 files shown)


Half-Shot:
>Question: by default @room only highlights if you are at least level 50 (moderator), what do?

We should probably find a way to start syncing roles over.

Sorunome:
> We should probably find a way to start syncing roles over.

OK, that would be a different, separate PR, though

src/messageprocessor.ts
```diff
@@ -171,6 +176,10 @@ export class MessageProcessor {
         return `${FLAG}${name}${FLAG}${node.animated ? 1 : 0}${FLAG}${node.id}${FLAG}`;
     }
 
+    public InsertRoom(): string {
+        return "@room";
```
Half-Shot: lol

src/messageprocessor.ts
```diff
@@ -213,6 +222,8 @@ export class MessageProcessor {
         return {
             channel: (node) => this.InsertChannel(node, msg),
             emoji: (node) => this.InsertEmoji(node),
+            everyone: (_) => this.InsertRoom(),
```
Half-Shot: *Surely* this could just be () => "@room"?

Sorunome:
This PR is for D->M parsing, and yes, @everyone and @here will look like @room on matrix. Currently it doesn't check if everyone / here was pinged, though that should be easily addable. Hang on juuuuust a sec!

package.json
```diff
@@ -38,7 +38,7 @@
     "bluebird": "^3.5.1",
     "command-line-args": "^4.0.1",
     "command-line-usage": "^4.1.0",
-    "discord-markdown": "git+https://github.com/Sorunome/discord-markdown.git",
+    "discord-markdown": "^2.0.0",
```
Half-Shot: ooh, did this get merged?

Half-Shot:
Good nuff for this :dog: . Merging

Issue #219, opened 2018-08-01
Changing avatar or username joins user


Sorunome:
If there is a room in a guild (discord side) that a user can't view (due to e.g. roles) if they change their username or their avatar their ghost will still be joined to the matrix-side room.

Half-Shot:
I suspect the user synchroniser fetches all channels in the guild and joins all of the users. We should be able to fix this by first checking channel by channel for guild state changes and then adding in code to leave them from rooms they no longer have permission to view.

Sorunome:
Yep, I was waiting on #194 first before addressing this

Sorunome:
merged

Sorunome:
merged

Issue #215, opened 2018-07-30
Handle role-based ghost joins/leavs


Sorunome:
When people get a role in a discord guild other rooms are visible for them. Joing / leaving a ghost in a matrix room based on that doesn't seem to be handled currently

Sorunome:
merged

PR #322, opened 2018-11-24
Update usersynchroniser to handle more things

Sorunome:
Updates the custom `m.room.member` key on e.g. role changes (can't find an issue for this atm)

Only join rooms you are in  #219 

Handle role-based join / leaves #215

Test failing is, again, due to that one PR that will fix things

### DIFF
--- a/src/bot.ts
+++ b/src/bot.ts
@@ -159,8 +159,8 @@ export class DiscordBot {
         client.on("userUpdate", async (_, user) => this.userSync.OnUpdateUser(user));
         client.on("guildMemberAdd", async (user) => this.userSync.OnAddGuildMember(user));
         client.on("guildMemberRemove", async (user) =>  this.userSync.OnRemoveGuildMember(user));
-        client.on("guildMemberUpdate", async (oldUser, newUser) =>
-            this.userSync.OnUpdateGuildMember(oldUser, newUser));
+        client.on("guildMemberUpdate", async (_, member) =>
+            this.userSync.OnUpdateGuildMember(member));
         client.on("debug", (msg) => { jsLog.verbose(msg); });
         client.on("error", (msg) => { jsLog.error(msg); });
         client.on("warn", (msg) => { jsLog.warn(msg); });
--- a/src/bot.ts
+++ b/src/bot.ts
@@ -439,10 +439,12 @@ export class DiscordBot {
         if (member) {
             let rooms: string[] = [];
             await Util.AsyncForEach(guild.channels.array(), async (channel) => {
-                if (!channel.members.has(member.id)) {
+                if (channel.type !== "text" || !channel.members.has(member.id)) {
                     return;
                 }
-                rooms = rooms.concat(await this.channelSync.GetRoomIdsFromChannel(channel));
+                try {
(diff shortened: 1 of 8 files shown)


src/usersyncroniser.ts
```diff
@@ -372,4 +390,14 @@ export class UserSyncroniser {
...
     }
+
+    private async leave(intent: Intent, roomId: string, checkCache: boolean = true) {
+        const userId = intent.getClient().getUserId();
+        if (checkCache && ![null, "join", "invite"]
+            .includes(intent.opts.backingStore.getMembership(roomId, userId))) {
+            return;
+        }
+        await intent.leave(roomId);
+        intent.opts.backingStore.setMembership(roomId, userId, "leave");
```
Half-Shot: I thought intent.leave would set the membership anyway, or the leave event that comes down the AS stream.

src/usersyncroniser.ts
```diff
@@ -372,4 +390,14 @@ export class UserSyncroniser {
     private displayNameForUser(discordUser): string {
         return `${discordUser.username}#${discordUser.discriminator}`;
     }
+
+    private async leave(intent: Intent, roomId: string, checkCache: boolean = true) {
+        const userId = intent.getClient().getUserId();
+        if (checkCache && ![null, "join", "invite"]
```
Half-Shot: We don't need to do this because a .leave should NO-OP quickly in synapse if the user isn't joined.

src/bot.ts
```diff
@@ -435,15 +435,32 @@ export class DiscordBot {
...
-            throw new Error("Room(s) not found.");
+    public async GetRoomIdsFromGuild(guild: Discord.Guild, member?: Discord.GuildMember): Promise<string[]> {
+        if (member) {
+            let rooms: string[] = [];
+            await Util.AsyncForEach(guild.channels.array(), async (channel) => {
+                if (channel.type !== "text" || !channel.members.has(member.id)) {
+                    return;
+                }
+                try {
+                    rooms = rooms.concat(await this.channelSync.GetRoomIdsFromChannel(channel));
```
Half-Shot: Wait, this can throw?

src/bot.ts
```diff
@@ -435,15 +435,32 @@ export class DiscordBot {
...
+                } catch (e) { } // no bridged rooms for this channel
+            });
+            if (rooms.length === 0) {
+                log.verbose(`Couldn't find room(s) for guild id:${guild.id} with member id:${member.id}.`);
+                throw new Error("Room(s) not found.");
```
Half-Shot: Possibly "no rooms were found for this guild" might be more correct?

src/usersyncroniser.ts
Sorunome: soru tested and apparently it doesn't

src/usersyncroniser.ts
Sorunome: Also tested this, and it doesn't do that, it keeps on sending leave events to synapse

src/bot.ts
Sorunome: yes, if there are no rooms in `this.channelSync.GetRoomIdsFromChannel`

src/usersyncroniser.ts
Half-Shot: ugh

PR #327, opened 2018-11-30
Matrix event handling restructure

Sorunome:
This event re-structures matrix event handling a bit to make more sense, code-flow wise.

It requires #322 and #316 to be merged

it introduces `matrixcommandhandler.ts` which is responsible for handling all the commands on the matrix side of things and moves functions around to their appropriate class so that they make more sense. (like, why did `bot.ts` and `matrixroomhandler.ts` handle matrix events if we have `matrixeventprocessor.ts`?)

Changes are starting 548019e


Half-Shot:
*looks slightly terrified*

Sorunome:
This PR was kinda rushed and should have proper discussion beforehand

Issue #335, opened 2018-12-10
Have replies tag the user


Sorunome:
in discord embeds you an do `[](<@userid>)` to silently ping people. Suggestion by devie

Half-Shot:
Do we want to let users be silently pinged outside of intentional use cases btw?

Sorunome:
What do you have in mind? M->D parsing to filter out the discord regexes to other stuff, like inserting a zero-width space to prevent pings? we could do that

Sorunome:
it appears that discord embeds don't ever ping the user no matter what --> this is impossible to do due to discord being discord

Sorunome:
added

Issue #329, opened 2018-12-02
M->D reply date


Sorunome:
add a footer tho reply embeds that says the original time. Suggestion by devie

Sorunome:
added

PR #350, opened 2018-12-24
add reply date to embed replies and ping discord users

Sorunome:
implements #329 and #335


Half-Shot:
Looks good, just needs a test ;)

PR #361, opened 2018-12-28
Send read receipts for successfully bridged Matrix events

### LINKED ISSUE #326: Use Matrix AS bot read recipient to mark sent status
make it so that the AS bot reads the event after transmitting it M->D

Johni0702:
Not sure what the policy on unit tests is but there weren't any for the two methods I changed, so I didn't add any.

Closes #326


Sorunome:
looks good, thank you!

Issue #364, opened 2019-01-05
"Timed out waiting for a response from the Discord owners" kills the bridge process


turt2live:
This is not fun :(

Half-Shot:
That shouldn't happen D:, logging?

turt2live:
for future explorers:
```
Jan-5 20:05:13.474 [Provisioner] info: Received response for bridging request in 531198884651728918 : responded=false approved=false expired=true
/build/src/provisioner.js:54
                        throw new Error("Timed out waiting for a response from the Discord owners");
                        ^

Error: Timed out waiting for a response from the Discord owners
    at approveFn (/build/src/provisioner.js:54:31)
    at Timeout.setTimeout [as _onTimeout] (/build/src/provisioner.js:62:30)
    at listOnTimeout (timers.js:324:15)
    at processTimers (timers.js:268:5)
```

(troubleshooted in the Matrix room)

PR #366, opened 2019-01-05
Fix provisioner not waiting

Half-Shot:


Half-Shot:
This might even fix #364, but will fix a security bug

PR #370, opened 2019-01-06
fix confusing port thing in readme

Sorunome:
Quite a few people had on the homeserver URL the `8448` port due to the readme stating that, however due to that commonly being a self-signed port stuff fails. This PR clarifies that


Half-Shot:
Hmm, but by default not including a port will land you on port 443, which isn't enabled by default?

Sorunome:
While that is true, by default you have an invalid ssl cert on port 8448, which also yields the bridge to not function.

As such, most proxy port 443 via e.g. nginx with a valid cert (to also make riot login easy), and thus that should cover most cases here.

Half-Shot:
I think that "(e.g. what is commonly found on port `8448`)" is excessive. We can just tell users to use a valid cert. It may confuse people more if we label 8448 as problematic as it's a red herring.

Half-Shot:
LGTM :)

Issue #377, opened 2019-01-18
Refactor - 2019 edition


Half-Shot:
- [x] Move user/room store to a postgres format (not sure yet whether do to at this level, or bridge lib level).
  - [x] Rooms
  - [x] Users
- [x] `Bluebird` should be obliterated, and we should find some util functions to replace it. #448 
- [x] Simplify M->D sending code.
- [x] Provisioner code should be taken out `bot.ts`, and use @Sorunome command code.
- [x] Matrix commands to use soru's command code. (and some of the discord stuff)
- [x] Refactor store function names. https://github.com/Half-Shot/matrix-appservice-discord/pull/455
- [x] ~Replace some funcs with idbdata functions.~ The only ones left are the user_id <=> discord_id+token mappings, which are quite complicated to merge into the idbdata format. I'm happy to leave this for another time, as it's not really used much anyway.
- [x] ~Remove `log.ts`~ Let's not do this now, since the move to the bot-sdk in #454 will leave us with no ability to log.
- [x] Take a look at circular dependencies in `discordas.ts` initialization logic.
- [x] `    "getbotlink": "node ./build/tools/addbot.js",` weird name.
- [x] Remove `getSchemaVersion` #391 
- [x]  `bot.ts` should not be at the center of processing events, factor out this.

Half-Shot:
Did #378 actually enforce " bot.ts should not be at the center of processing events, factor out this." further?

Sorunome:
Fully done! :tada:

PR #378, opened 2019-01-22
Refactor circular dependencies

Sorunome:
This removes circular dependency stuff, as per #377

### DIFF
--- a/src/bot.ts
+++ b/src/bot.ts
@@ -62,11 +62,9 @@ interface IThirdPartyLookup {
 }
 
 export class DiscordBot {
-    private config: DiscordBridgeConfig;
     private clientFactory: DiscordClientFactory;
     private store: DiscordStore;
     private bot: Discord.Client;
-    private bridge: Bridge;
     private presenceInterval: number;
     private sentMessages: string[];
     private lastEventIds: { [channelId: string]: string };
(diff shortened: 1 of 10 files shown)


src/discordas.ts
```diff
@@ -122,17 +128,29 @@ async function run(port: number, fileConfig: DiscordBridgeConfig) {
...
-    provisioner.SetBridge(bridge);
-    roomhandler.setBridge(bridge);
-    discordbot.setBridge(bridge);
-    discordbot.setRoomHandler(roomhandler);
+    // Warn and deprecate old config options.
+    const discordbot = new DiscordBot(botUserId, config, bridge);
+    const roomhandler = discordbot.RoomHandler;
+
+    try {
+        callbacks.onAliasQueried = roomhandler.OnAliasQueried.bind(roomhandler);
```
Half-Shot: Clever!

src/bot.ts
```diff
@@ -76,33 +76,36 @@ export class DiscordBot {
...
     private channelSync: ChannelSyncroniser;
     private roomHandler: MatrixRoomHandler;
+    private provisioner: Provisioner;
+    private botUserId: string;
 
     /* Handles messages queued up to be sent to discord. */
     private discordMessageQueue: { [channelId: string]: Promise<void> };
 
-    constructor(config: DiscordBridgeConfig, store: DiscordStore, private provisioner: Provisioner) {
+    constructor(botUserId: string, config: DiscordBridgeConfig, bridge: Bridge) {
```
Half-Shot:
```suggestion
    constructor(private botUserId: string, config: DiscordBridgeConfig, private bridge: Bridge) {
```

Sorunome: uuuuh, they are set right after to the private vars?

Currently there is a mixture when setting private vars passed to the constructor which way it should be done, so, if soru changes this she should go over the other ones and change it there, too.

Sooo... TL;DR which syntax should we use? the one you suggested?

Half-Shot: I kiiiinda prefer the syntax suggested. The reason it wasn't done is because I don't believe it was supported when we started the project (or I didn't notice, one of them). I very much prefer saving us two lines :)

Issue #321, opened 2018-11-23
Move User/Room stores to using our Storage layer


Sorunome:
use sql instead of user-store / room-store. This allows for better performance, as we have a postgresql option

Sorunome:
Done in #383 and #394

Issue #225, opened 2018-08-06
Blank room store entires mess stuff up


Sorunome:
```
{"id":"!redacted:domain.com","matrix_id":"!redacted:domain.com","matrix":{"extras":{}},"_id":"S0jvSr7psDo6IDNa"}
{"id":"!redacted:domain.com   discord_redacted_redacted_bridged","remote_id":"discord_redacted_redacted_bridged","matrix_id":"!redacted:domain.com","remote":{"discord_type":"text","discord_guild":"redacted","discord_channel":"redacted","plumbed":true},"matrix":{"extras":{}},"data":{},"_id":"s2DDeczYDNFpn8uc"}
```

Sorunome:
This should be fixed in #383

PR #383, opened 2019-01-27
Use database as a roomstore

Half-Shot:
This replaces the room store component of the appservice-bridge lib. It uses the same function signatures where possible, but be warned that it only implements what is strictly necessary.

Things to do:
* ~Fix broken tools.~
* ~Fix broken tests.~
* Cache more?
* User store?

(This resolves some of #321 )


package.json
```diff
@@ -14,7 +14,8 @@
     "usertool": "node ./build/tools/userClientTools.js",
     "directoryfix": "node ./build/tools/addRoomsToDirectory.js",
     "ghostfix": "node ./build/tools/ghostfix.js",
-    "chanfix": "node ./build/tools/chanfix.js"
+    "chanfix": "node ./build/tools/chanfix.js",
+    "moveRoomStore": "node ./build/tools/moveRoomStoreToDb.js"
```
Sorunome: Is this tool needed if we have stuff in the migraiton?

src/bot.ts
```diff
@@ -500,9 +500,9 @@ export class DiscordBot {
             throw Error("Room(s) not found.");
         }
         const entry = entries[0];
-        const guild = client.guilds.get(entry.remote.get("discord_guild"));
+        const guild = client.guilds.get(entry.remote!.get("discord_guild") as string);
```
Sorunome: why are we assuming `entry.remote` is defined? shouldn't we check for this somehow?

Same goes for all the other additions of `!`

src/db/connector.ts
```diff
@@ -14,12 +14,14 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+type SQLTYPES = number | boolean | string | null;
```
Sorunome: Nice, good thinking!

src/db/roomstore.ts
```diff
@@ -0,0 +1,347 @@
...
+
+interface IRemoteRoomData extends IRemoteRoomDataLazy {
+    discord_guild: string;
+    discord_channel: string;
+}
+
+interface IRemoteRoomDataLazy  {
+    discord_guild?: string;
+    discord_channel?: string;
+    discord_name?: string|null;
```
Sorunome: why do we have both `?` (implying `undefined`) and `null` on these?

src/db/roomstore.ts
```diff
@@ -0,0 +1,347 @@
...
+    update_icon?: number|boolean|null;
+    plumbed?: number|boolean|null;
+}
+
+export class RemoteStoreRoom {
+    public data: IRemoteRoomData;
+    constructor(public readonly roomId: string, data: IRemoteRoomData) {
+        for (const k of ["discord_guild", "discord_channel", "discord_name",
+        "discord_topic", "discord_iconurl", "discord_iconurl_mxc", "discord_type"]) {
+            data[k] = typeof(data[k]) === "number" ? String(data[k]) : data[k] || null;
```
Sorunome: if `data[k]` doesn't exist, shouldn't we rather set `undefined` than `null`?

src/db/roomstore.ts
```diff
@@ -0,0 +1,347 @@
...
+        }
+        for (const k of ["update_name", "update_topic", "update_icon", "plumbed"]) {
+            data[k] = Number(data[k]) || 0;
```
Sorunome: These are technically booleans, why do we treat them as numbers?

src/db/roomstore.ts
```diff
@@ -0,0 +1,347 @@
...
+
+    public getId() {
+        return this.roomId;
+    }
+}
+
+export interface IRoomStoreEntry {
+    id: string;
+    matrix: MatrixStoreRoom|null;
+    remote: RemoteStoreRoom|null;
```
Sorunome: won't some things break if we leave out the shared `data` key, even if we don't use it? Same for `matrix_id` and `remote_id`

src/db/roomstore.ts
```diff
@@ -0,0 +1,347 @@
...
+    }
+
+    public async upsertEntry(entry: IRoomStoreEntry) {
+        const promises: Promise<void>[] = [];
+
+        const row = (await this.db.Get("SELECT * FROM room_entries WHERE id = $id", {id: entry.id})) || {};
+
+        if (!row.id) {
+            // Doesn't exist at all, create the room_entries row.
+            const values = {
```
Sorunome: `const values: IRoomStoreEntry` (missing type)

src/db/roomstore.ts
```diff
@@ -0,0 +1,347 @@
...
+            try {
+                await this.db.Run(`INSERT INTO room_entries VALUES ($id, $matrix, $remote)`, values);
+                log.verbose("Created new entry " + entry.id);
+            } catch (ex) {
+                log.error("Failed to insert room entry", ex);
+                throw Error("Failed to insert room entry");
+            }
+        }
+
+        const matrixId = entry.matrix ? entry.matrix.roomId : null;
```
Sorunome: shouldn't this stuff go into an else-block? we unneededly check if we need updating right after we inserted a new entry

src/db/roomstore.ts
```diff
@@ -0,0 +1,347 @@
...
+        const remoteId = entry.remote ? entry.remote.roomId : null;
+        const mxIdDifferent = matrixId !== row.matrix_id;
+        const rmIdDifferent = remoteId !== row.remote_id;
+        // Did the room ids change?
+        if (row.id && (mxIdDifferent || rmIdDifferent)) {
```
Sorunome: yes, we check again for `row.id` here, but it is kinda....redundant, we do the same check twice

src/db/roomstore.ts
```diff
@@ -0,0 +1,347 @@
...
+            "SELECT * FROM room_entries WHERE matrix_id = $id", {id: matrixId},
+        );
+        const res: IRoomStoreEntry[] = [];
+        for (const entry of entries) {
+            const remoteId = entry.remote_id as string || "";
+            const row = await this.db.Get(
+                "SELECT * FROM remote_room_data WHERE room_id = $rid",
+                {rid: remoteId},
+            );
+            if (!row) { continue; }
```
Sorunome: why continue? shouldn't we push a thing with blank remote data?

src/db/roomstore.ts
```diff
@@ -0,0 +1,347 @@
...
+
+            res.push({
+                id: (entry.id as string),
+                matrix: matrixId ? new MatrixStoreRoom(matrixId) : null,
+                // tslint:disable-next-line no-any
+                remote: remoteId ? new RemoteStoreRoom(remoteId, row as any) : null,
```
Sorunome: why `any` and not `IRemoteRoomDataLazy`

src/db/roomstore.ts
```diff
@@ -0,0 +1,347 @@
...
+        }
+        if (res.length > 0) {
+            this.entriesMatrixIdCache.set(matrixId, {e: res, ts: Date.now()});
+        }
+        return res;
+    }
+
+    public async getEntriesByMatrixIds(matrixIds: string[]): Promise<IRoomStoreEntry[]> {
+        const entries = await this.db.All(
+            `SELECT * FROM room_entries WHERE matrix_id IN ('${matrixIds.join("','")}')`,
```
Sorunome: this doesn't look particularly SQL injection safe. One could easily make a homeserver which creates sql-injection-ready room IDs and poof goes travis' DB

src/db/roomstore.ts
```diff
@@ -0,0 +1,347 @@
...
+        );
+        const res: IRoomStoreEntry[] = [];
+        for (const entry of entries) {
+            const matrixId = entry.matrix_id as string || "";
+            const remoteId = entry.remote_id as string || "";
+            const row = await this.db.Get(
+                "SELECT * FROM remote_room_data WHERE room_id = $rid",
+                {rid: remoteId},
+            );
+            if (!row) { continue; }
```
Sorunome: same, why not append blank?

src/db/roomstore.ts
```diff
@@ -0,0 +1,347 @@
...
+                remote: remoteId ? new RemoteStoreRoom(remoteId, row as any) : null,
+            });
+        }
+        if (res.length > 0) {
+            this.entriesMatrixIdCache.set(matrixId, {e: res, ts: Date.now()});
+        }
+        return res;
+    }
+
+    public async getEntriesByMatrixIds(matrixIds: string[]): Promise<IRoomStoreEntry[]> {
```
Sorunome: This function doesn't cache any?

src/db/roomstore.ts
```diff
@@ -0,0 +1,347 @@
...
+            log.error("Failed to insert room entry", ex);
+            throw Error("Failed to insert room entry");
+        }
+    }
+
+    public async setMatrixRoom(matrixRoom: MatrixStoreRoom) {
+        // This no-ops, because we don't store anything interesting.
+    }
+
+    public async getEntriesByRemoteRoomData(data: IRemoteRoomDataLazy): Promise<IRoomStoreEntry[]> {
```
Sorunome: this whole method feels clunky

src/db/roomstore.ts
```diff
@@ -0,0 +1,347 @@
...
+    }
+
+    public async removeEntriesByRemoteRoomId(remoteId: string) {
+        await this.db.Run(`DELETE FROM room_entries WHERE remote_id = $remoteId`, {remoteId});
+        await this.db.Run(`DELETE FROM remote_room_data WHERE room_id = $remoteId`, {remoteId});
+    }
+
+    public async removeEntriesByMatrixRoomId(matrixId: string) {
+        await this.db.Run(`DELETE FROM room_entries WHERE matrix_id = $matrixId`, {matrixId});
+        await this.db.Run(`DELETE FROM remote_room_data WHERE room_id = $matrixId`, {matrixId});
```
Sorunome: eeer this wouldn't find anything? we'd need to select the remote room_id first before deleting it from room_entries?

src/db/roomstore.ts
```diff
@@ -0,0 +1,347 @@
...
+    }
+
+    private async upsertRoom(room: RemoteStoreRoom) {
+        const existingRow = await this.db.Get(
+            "SELECT * FROM remote_room_data WHERE room_id = $id",
+            {id: room.roomId},
+        );
+
+        if (!room.data) {
+            throw new Error("Tried to upsert a room with undefined data");
```
Sorunome: we could check for this before performing a db query!

src/db/roomstore.ts
```diff
@@ -0,0 +1,347 @@
...
+
+        const data = {
+            discord_channel:     room.data.discord_channel,
+            discord_guild:       room.data.discord_guild,
+            discord_iconurl:     room.data.discord_iconurl,
+            discord_iconurl_mxc: room.data.discord_iconurl_mxc,
+            discord_name:        room.data.discord_name,
+            discord_topic:       room.data.discord_topic,
+            discord_type:        room.data.discord_type,
+            plumbed:             room.data.plumbed || 0,
```
Sorunome: shouldn't these be booleans?

src/db/roomstore.ts
```diff
@@ -0,0 +1,347 @@
...
+            `,
+            {
+                id: room.roomId,
+                // tslint:disable-next-line no-any
+                ...data as any,
+            });
+            return;
+        }
+
+        const keysToUpdate = { };
```
Sorunome: this doesn't have a type

src/db/schema/v8.ts
```diff
@@ -0,0 +1,87 @@
...
+                await store.roomStore.linkRooms(matrix, remote);
+                log.info(`Migrated ${matrix.roomId}`);
+            } catch (ex) {
+                log.error(`Failed to link ${matrix.roomId}: `, ex);
+            }
+        }
+    }
+
+    public async rollBack(store: DiscordStore): Promise<void> {
+        await store.db.Run(
```
Sorunome: do we want to restore old roomstore on rollback?

src/discordas.ts
```diff
@@ -94,8 +93,17 @@ async function run(port: number, fileConfig: DiscordBridgeConfig) {
                     return await roomhandler.OnAliasQuery.bind(roomhandler)(alias, aliasLocalpart);
                 } catch (err) { log.error("Exception thrown while handling \"onAliasQuery\" event", err); }
             },
-            onEvent: async (request, context) => {
+            onEvent: async (request) => {
                 try {
+                    // Build our own context.
```
Sorunome: why do we need this?

src/matrixroomhandler.ts
```diff
@@ -83,14 +86,28 @@ export class MatrixRoomHandler {
...
+    public setBridge(bridge: Bridge, roomStore: DbRoomStore) {
         this.bridge = bridge;
+        this.roomStore = roomStore;
     }
 
     public async OnAliasQueried(alias: string, roomId: string) {
         log.verbose(`Got OnAliasQueried for ${alias} ${roomId}`);
         let channel: Discord.GuildChannel;
         try {
+            // We previously stored the room as an alias.
```
Sorunome: why not migrate this in the migration?

src/provisioner.ts
```diff
@@ -20,32 +20,32 @@ import {
...
 
 export class Provisioner {
 
-    private bridge: Bridge;
     private pendingRequests: Map<string, (approved: boolean) => void> = new Map(); // [channelId]: resolver fn
+    private roomStore: DbRoomStore;
 
-    public SetBridge(bridge: Bridge): void {
-        this.bridge = bridge;
+    public setStore(roomStore: DbRoomStore) {
```
Sorunome: can't this go in the constructor? with #378 maybe?

tools/addRoomsToDirectory.ts
```diff
@@ -89,20 +90,20 @@ const bridge = new Bridge({
...
     registration: true,
-    roomStore: options.store,
 });
 
+const discordstore = new DiscordStore(config.database ? config.database.filename : "discord.db");
+
 async function run() {
     try {
-        await bridge.loadDatabases();
+        await discordstore.init();
```
Sorunome: with #378 this should go into that discorbot.init

package.json
Half-Shot: No, and I deleted it. I should also remove it from package.json

src/db/roomstore.ts
Half-Shot: It's a confusing mess :/. We need these to be undefined so that when you want to search for a val, you can leave most of these undefined. However, you may also want to set it to null to explicitly look for null values (since SQLITE uses null to determine an empty value).

Tl;dr - undefined == don't search for this. null == look for a empty value.

src/db/roomstore.ts
Half-Shot: Nope, because undefined doesn't exist in SQL terms but null means empty.

src/db/roomstore.ts
Half-Shot: SQLITE has no bool type, so we have to do conversions :/

src/db/roomstore.ts
Half-Shot: Indeed, this PR is doing the bare minimum to make everything work. I'd expect us to implement matrix and link data when it becomes required. Obviously typescript won't let us just set values without any typing information, anyway.

src/db/roomstore.ts
Half-Shot: Maybe, but I think this was a tradeoff between a few nano seconds and less indent/less ugly code.

src/db/roomstore.ts
Half-Shot: oh derp

src/db/roomstore.ts
Half-Shot: oh um...probably

src/db/roomstore.ts
Half-Shot: I suspect I forgot about it.

src/db/roomstore.ts
Half-Shot: Oh balls, yes you are right (though I hoped that sqlite3 / pg might be doing minimal escaping here).

We should validate that the matrixId really is an id (although I use aliases briefly in MRH).

src/db/roomstore.ts
Half-Shot: I don't think we use it in a high traffic area, so I avoided over optimizing it.

src/db/roomstore.ts
Half-Shot: I don't think it's that bad? We build up a WHERE clause based on provided data and iterate over the result. Suggestions welcome though, since I might be missing something.

src/db/roomstore.ts
Half-Shot: GAH, yes. Probably sensible to do a SELECT, then two deletes.

src/db/roomstore.ts
Half-Shot: Same as aboove.

src/db/roomstore.ts
Half-Shot: I suspect I was being lazy, this should probably be `IWhatItsNameLazy`

src/db/schema/v8.ts
Half-Shot: Well, we don't delete any values from the roomstore. The bridge won't function without it though, and trying to write code to if else around the room stores is going to be a nightmare.

src/db/roomstore.ts
Sorunome: Or we could create an array of parameters `WHERE matrix_id IN ($1, $2, $3, ....)` and then pass those on

src/discordas.ts
Half-Shot: We used to use the nedb context in a few places to find out which remote room we were handling, now we do the same thing with the db.

src/matrixroomhandler.ts
Half-Shot: huh? No, we have to store the matrix_id as an alias in `createMatrixRoom` because we don't know the roomId yet.

src/db/roomstore.ts
Sorunome: not really sure, it just... *feels* clunkly and hard to cache if something happens down the road. Probably fine

src/provisioner.ts
Half-Shot: Probably with #378 :)

tools/addRoomsToDirectory.ts
Half-Shot: Ditto!

src/db/roomstore.ts
Half-Shot: It's not, as it turns out because we are inserting this into the DB which wants strings.

src/db/roomstore.ts
```diff
@@ -0,0 +1,365 @@
+/*
+Copyright 2018 matrix-appservice-discord
```
turt2live:
```suggestion
Copyright 2019 matrix-appservice-discord
```

src/db/schema/v8.ts
```diff
@@ -0,0 +1,87 @@
+/*
+Copyright 2018 matrix-appservice-discord
```
turt2live:
```suggestion
Copyright 2019 matrix-appservice-discord
```

test/db/test_roomstore.ts
```diff
@@ -0,0 +1,195 @@
+/*
+Copyright 2017, 2018 matrix-appservice-discord
```
turt2live:
```suggestion
Copyright 2019 matrix-appservice-discord
```

src/db/roomstore.ts
```diff
@@ -0,0 +1,365 @@
...
+        const res: IRoomStoreEntry[] = [];
+        for (const entry of entries) {
+            let remote: RemoteStoreRoom|null = null;
+            if (entry.remote_id) {
+                const remoteId = entry.remote_id as string;
+                const row = await this.db.Get(
+                    "SELECT * FROM remote_room_data WHERE room_id = $remoteId",
+                    {remoteId},
+                );
+                if (row) {
```
Sorunome: actually, thinking about it more it might be a good idea to remove the entries without a remote. That should kill #225 once and for all as it *should* always be present

src/db/roomstore.ts
```diff
@@ -0,0 +1,365 @@
...
+        for (const entry of entries) {
+            let remote: RemoteStoreRoom|null = null;
+            const matrixId = entry.matrix_id as string || "";
+            const remoteId = entry.remote_id as string;
+            if (remoteId) {
+                const row = await this.db.Get(
+                    "SELECT * FROM remote_room_data WHERE room_id = $rid",
+                    {rid: remoteId},
+                );
+                if (row) {
```
Sorunome: same here, maybe it is actually a good idea to ignore results without remote data (like you did originally)

tools/addRoomsToDirectory.ts
Sorunome: also this still

PR #391, opened 2019-02-04
As per #377, we don't use this anymore

Half-Shot:


Sorunome:
what should soru even review on this PR :thinking:

Half-Shot:
idk, can I have a :tada: :P

PR #409, opened 2019-02-26
Point "new application" link a bit higher up since it doesn't work anymore

keithzg:
Following the link for creating a new application actually results in

> Application not found!
> There was no application found with ID create.

This error page has no clear way to actually create an application. Going up to https://discordapp.com/developers/applications/, there's at least a button in the top-right for "New Application".

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,7 +85,7 @@ should show up in the network list on Riot and other clients.
 
 ### Setting up Discord
 
-* Create a new application via https://discordapp.com/developers/applications/me/create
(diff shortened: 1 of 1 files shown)


Sorunome:
Thank you for your PR! It looks good, however, would you mind pointing it to the `develop` branch for now? Thank you!

keithzg:
Whoops, sorry! I had just done it all very lazily against the branch Github defaulted to ;)

PR #394, opened 2019-02-09
Hs/userstore db

Half-Shot:
This depends on #383.

Currently missing tests and fixes to the tools.

PR #417, opened 2019-03-23
Remove extraneous comma on UPDATE

turt2live:
A line feed is also removed to ensure the space doesn't get washed away by accident.

Fixes https://github.com/Half-Shot/matrix-appservice-discord/pull/394#issuecomment-464302619

**Note the target branch: This is against the user store DB PR**


Half-Shot:
Gracias

PR #419, opened 2019-03-24
Update docs/howto.md for internal room id

erdnaxe:
Update docs/howto.md to let new users know that they should use the internal
room id when using ``adminme`` script.


docs/howto.md
```diff
@@ -18,6 +18,9 @@ is formatted as https://discordapp.com/channels/``guildid``/``channelid``
   * ``npm run adminme -- -r '!AbcdefghijklmnopqR:example.com' -u '@Alice:example.com' -p '100'``
   * Run ``npm run adminme -- -h`` for usage.
 
+Please note that ``'!AbcdefghijklmnopqR:example.com'`` is the internal room id and will always begin with ``!``.
```
Half-Shot: You can use single backticks here, and you don't need the `'`s

Half-Shot:
This definitely clears up confusion, thanks for spotting this.

PR #420, opened 2019-03-26
[D->M] Parses embed.fields for discord embeds

Sorunome:


test/test_discordmessageprocessor.ts
```diff
@@ -537,6 +537,38 @@ describe("DiscordMessageProcessor", () => {
...
+            const inContent = "Content that goes in the message";
+            const content = processor.InsertEmbeds(inContent, msg);
+            Chai.assert.equal(
+                content,
+`Content that goes in the message
+
+----
+##### [TestTitle](testurl)
+TestDescription
+**fox**
```
Half-Shot: Hmm, do we want these as a table, or at least next to each other rather than on newlines. E.g. `**fox**: floof` ?

Sorunome: discord actually has an option for that, the `inline` property https://discord.js.org/#/docs/main/stable/class/MessageEmbedField

For simplicity soru assumed that inline is false

Half-Shot: Can haz screenshot? :3

Sorunome: Looks a bit silly because silly testing bot. No actual message content, thus the horizontal bar as start of embed

Screenshot_2019-04-23_21-20-01

Half-Shot: I guess that works :)

Half-Shot:
Merge merge merge

Issue #425, opened 2019-04-10
Documentation on bot roles in Discord that may stop a bridge


serra-allgood:
Had a hell of a time figuring out why the bot was working on a fresh Discord server but not on one that I'm on all the time. It had to do with how roles are configured on the server and how bot roles are placed in the role hierarchy for a guild.

1. Bot roles always start at the bottom of the roles
2. Although you can assign any role to a bot, only roles that are lower in hierarchy (which is higher in position in the UI) will actually be applied to the bot in terms of permissions. So if you have a channel that prevents anyone from posting except for members with a certain role, this will prevent the bot from posting in that channel.
3. If the `@everyone` role is configured on individual channels to outright deny the `Manage Webhooks` permission, it will prevent the bot both from being able to create webhooks for the channel and post in the channel even if a hook exists.

Perhaps this is all well known to Discord devs, but it was very opaque to me and required digging into the API documentation to fully understand. Additionally, because I'm not a mod of the server I'm trying to bridge to, it took some raw API calls in Postman using the bot token for me to piece everything together.

I should be able to put together a PR for this documentation if it's deemed worthy of inclusion.

Half-Shot:
To be honest, I'm not terribly sure how the role system works myself. It would be extremely useful if this was noted down in a markdown doc and PRed :)

serra-allgood:
Turns out I was mistaken, so closing this issue

PR #426, opened 2019-04-13
Create troubleshooting.md to explain role scenarios that may prevent the bridge from working

serra-allgood:
Addresses #425

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -1,0 +1,35 @@
+### Troubleshooting your Discord Bot
+
+So you've got the bridge running and the Discord bot invited to the
+server you want to bridge to, but the bridge isn't sending messages from
+Matrix to Discord? First, do a santiy check and create a fresh Discord server,
+invite the bot to it, and see if you can send messages back and forth. If it's
+working on a fresh server, you might be encountering issues with how permissions
+are calculated in Discord. Permissions for a given guild/channel/role are
+calculated as follows:
+
+1. Base permissions given to `@everyone` are applied at a guild level
+2. Permissions allowed to a user by their roles are applied at a guild level
+3. Overwrites that deny permissions for @everyone are applied at a channel level
+4. Overwrites that allow permissions for @everyone are applied at a channel level
+5. Overwrites that deny permissions for specific roles are applied at a channel level
+6. Overwrites that allow permissions for specific roles are applied at a channel level
+7. Member-specific overwrites that deny permissions are applied at a channel level
+8. Member-specific overwrites that allow permissions are applied at a channel level
+
+There are several scenarios that will prevent communication from Matrix to Discord.
+Here's what's been identified so far and how to fix them.
+
+- The `@everyone` role denies `Manage Webhooks` at the guild level.
+  Remove the denial.
+- In the channel, a certain role is required to send messages.
+  This is tricky. If the required role is applied to the bot, it may seem like
+  the bot has the role. However, the way roles work is that you only get the permissions
+  of the "lowest" role, and bots automatically start as the lowest role. In the roles
+  UI for a guild, roles are listed from lowest to highest. To be able to give the
+  permissions of a required role to a bot, the bot role must be lower in the list (which
+  means higher in position) then the required role.
+- The `@everyone` role denies `Manage Webhooks` at the channel level
+  Remove the denial.
+
+Other scenarios should be variations of these.
--- a/README.md
+++ b/README.md
@@ -103,6 +103,7 @@ should show up in the network list on Riot and other clients.
 * ``npm start``
 
 [Howto](./docs/howto.md)
+[Troubleshooting](./docs/troubleshooting.md)
 
 ## Features and Roadmap
 In a vague order of what is coming up next


Sorunome:
It is enough that the bot itself has all the webhook permissions, `@everyone` doesn't need it

serra-allgood:
Correct, but if everyone is marked as denying webhooks, it will prevent the bot from communicating from Matrix to Discord

Sorunome:
not if in advance permission system  it is set to allow for the bot role

Am 23. April 2019 01:15:09 MESZ schrieb Serra Allgood <notifications@github.com>:

serra-allgood:
I will double check but I'm almost certain that because `@everyone` will always be in a higher position than the bot role, if a specific channel is marked as deny webhooks for `@everyone`, it will break the bot as I described. I'll follow up tomorrow with a screen cast if I'm correct.

Sorunome:
> because `@everyone` will always be in a higher position than the bot role

That's....not really true at all
Screenshot_2019-04-23_18-35-12

serra-allgood:
Not to sound mean, but did you read the doc or was it unclear? Higher position, programmatically, means lower in the UI

Sorunome:
The bot role is not in a lower position than `@everyone`, in fact it is in a higher one

docs/troubleshooting.md
```diff
@@ -0,0 +1,35 @@
...
+4. Overwrites that allow permissions for @everyone are applied at a channel level
+5. Overwrites that deny permissions for specific roles are applied at a channel level
+6. Overwrites that allow permissions for specific roles are applied at a channel level
+7. Member-specific overwrites that deny permissions are applied at a channel level
+8. Member-specific overwrites that allow permissions are applied at a channel level
+
+There are several scenarios that will prevent communication from Matrix to Discord.
+Here's what's been identified so far and how to fix them.
+
+- The `@everyone` role denies `Manage Webhooks` at the guild level.
```
Sorunome: what soru means is like instead of removing the denial at `@everyone` you can give the bot role an override for allow anyways, something like that

docs/troubleshooting.md
```diff
@@ -0,0 +1,35 @@
...
+  Remove the denial.
+- In the channel, a certain role is required to send messages.
+  This is tricky. If the required role is applied to the bot, it may seem like
+  the bot has the role. However, the way roles work is that you only get the permissions
+  of the "lowest" role, and bots automatically start as the lowest role. In the roles
+  UI for a guild, roles are listed from lowest to highest. To be able to give the
+  permissions of a required role to a bot, the bot role must be lower in the list (which
+  means higher in position) then the required role.
+- The `@everyone` role denies `Manage Webhooks` at the channel level
```
Sorunome: ...or add an override for the bot role to allow it for them

docs/troubleshooting.md
```diff
@@ -0,0 +1,35 @@
...
+
+There are several scenarios that will prevent communication from Matrix to Discord.
+Here's what's been identified so far and how to fix them.
+
+- The `@everyone` role denies `Manage Webhooks` at the guild level.
+  Remove the denial.
+- In the channel, a certain role is required to send messages.
+  This is tricky. If the required role is applied to the bot, it may seem like
+  the bot has the role. However, the way roles work is that you only get the permissions
+  of the "lowest" role, and bots automatically start as the lowest role. In the roles
```
Sorunome: `@everyone` is still lower than the bot role

docs/troubleshooting.md
```diff
@@ -0,0 +1,35 @@
...
+  UI for a guild, roles are listed from lowest to highest. To be able to give the
```
Sorunome: the top role in the UI list is the one with most power, the lowest one with lowest power.

that is why admin roles are typically all on top, so that they can actually manage all roles below them

serra-allgood:
I don't really have time at this exact moment to respond at length. Like I said in my initial response, I will double check when I do have time. I remember going through the API docs and being confused as to why it was not working, and it was not until I created a fresh server, invited my bot, and played around with guild and channel permissions that I was able to debug. 

Perhaps there is another unknown confounding variable that affected what I literally experienced, and I assure you, did not hallucinate. If I cannot replicates, I will close this PR. If I can replicate, I will document the exact steps to do so with text and gif instructions. 

Also, I remember reading another part of the API docs that expressly say bots by default have the lowest position when added to a server. I confirmed as much through calling the API through a raw request UI. 

Please do not dismiss what I literally went through and quote API docs like I didn't read them. You already expressed doubt and I validated that by committing to going through a more rigorous process and providing demonstrable proof. 

Thank you.

Sorunome:
> I remember going through the API docs and being confused as to why it was not working

well, in the API the indexes for groups are so that `@everyone` is 0 and other groups have higher indexes. However, in the GUI the other roles are displayed on top, so it is sorted greatest-to-smallest. Maybe that causes some confusion?

> Also, I remember reading another part of the API docs that expressly say bots by default have the lowest position when added to a server.

lower than all other groups, except of `@everyone` as that one is special (as far as soru is aware)

Sparkfiber:
It should be noted that allowing the permission to manage webhooks for the ``@everyone`` role on the guild level is a **MAJOR** security risk for servers because it then allows anyone to create and use a webhook. This should **NEVER** be a setting on any public guild. Trust me - I have over 1800 members in my guild and this would be a massive security risk.  

Instead - you can use the role created when you invited the bot to the guild and give it the following permissions in the channels you wish to bridge. 
- Manage Webhooks
- Read Messages
- Send Messages
- Embed Links
- Attach Files
- Read Message History

On Discord, the channel-level permissions will trump guild-level permissions. Also @Sorunome is correct regarding the levels. ``@everyone`` is the default role in every server and is always at the bottom of the list. The roles above ``@everyone`` are technically a higher power. Note how the staff hierarchy is here:

hierarchy
If Admin was below Mod then the people with the Mod role could assign themselves admin privileges. 

I have set up the bridge and use the following permissions successfully:

**Guild-Level Role Permissions**
For the ``@everyone`` role:
Discord_H8IEI6b8iX
For the bot role (in this case LFG is the name of the bot):
Discord 2

**Channel-Level Role Permissions**
For the ``@everyone`` role:
Discord 3
For the bot role:
Discord 4

serra-allgood:
Another unknown must have been preventing communication, and I construed my explanation from frustration due to not having visibility into the role permissions of the guild I was trying to bridge. In my defense, I did not mean to instruct anyone to allow manage webhooks on `@everyone`, only to prevent it from being specifically denied in the case of channel permissions. I appreciate your explanation and detail, especially because I hadn't set aside time to properly investigate my understanding. 

Closing this PR as I was mistaken.

Issue #430, opened 2019-04-15
Show (Discord) behind "virtual" users on Matrix side


Perflyst:
All other bridges I now have such indicator, would be glad if you add one. I haven't found another issue to this, please close if this is a duplicate.

turt2live:
if this happens, please make it a config option. I very much like having it not there.

Sorunome:
Implemented in #440

PR #440, opened 2019-04-26
add name tags

Sorunome:
This implements #430


Sorunome:
Well, gonna add that, then! xP

Sorunome:
@Half-Shot is this better?

As for config testing, i couldn't see any specific config values being tested anywhere?

Half-Shot:
@Sorunome I may be getting my bridges confused, looks like we don't have that kind of test here. Ignore me.

Also, this looks fantastic. Thank you!

PR #448, opened 2019-04-29
Purge bluebird

Half-Shot:
Fixes part of #377

### DIFF
--- a/src/bot.ts
+++ b/src/bot.ts
@@ -34,7 +34,6 @@ import { ChannelSyncroniser } from "./channelsyncroniser";
 import { MatrixRoomHandler } from "./matrixroomhandler";
 import { Log } from "./log";
 import * as Discord from "discord.js";
-import * as Bluebird from "bluebird";
 import * as mime from "mime";
 import { IMatrixEvent, IMatrixMediaInfo } from "./matrixtypes";
 
@@ -186,11 +185,11 @@ export class DiscordBot {
         });
 
         // Due to messages often arriving before we get a response from the send call,
-        // messages get delayed from discord. We use Bluebird.delay to handle this.
+        // messages get delayed from discord. We use Util.Delay to handle this.
 
         client.on("messageDelete", async (msg: Discord.Message) => {
             try {
-                await Bluebird.delay(this.config.limits.discordSendDelay);
+                await Util.DelayedPromise(this.config.limits.discordSendDelay);
                 this.discordMessageQueue[msg.channel.id] = (async () => {
                     await (this.discordMessageQueue[msg.channel.id] || Promise.resolve());
                     try {
@@ -205,7 +204,7 @@ export class DiscordBot {
         });
         client.on("messageUpdate", async (oldMessage: Discord.Message, newMessage: Discord.Message) => {
             try {
-                await Bluebird.delay(this.config.limits.discordSendDelay);
+                await Util.DelayedPromise(this.config.limits.discordSendDelay);
                 this.discordMessageQueue[newMessage.channel.id] = (async () => {
                     await (this.discordMessageQueue[newMessage.channel.id] || Promise.resolve());
                     try {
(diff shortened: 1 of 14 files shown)


src/bot.ts
```diff
@@ -186,11 +185,11 @@ export class DiscordBot {
         });
 
         // Due to messages often arriving before we get a response from the send call,
-        // messages get delayed from discord. We use Bluebird.delay to handle this.
+        // messages get delayed from discord. We use Util.Delay to handle this.
```
Sorunome: what is it now? `Util.Delay` or `Util.DelayedPromise`? be consistent ^-^

src/clientfactory.ts
```diff
@@ -64,30 +62,28 @@ export class DiscordClientFactory {
...
-                resolve(id);
-            });
-            client.login(token).catch(reject);
-        }).timeout(READY_TIMEOUT).catch((err: Error) => {
-            log.warn("Could not login as a normal user.", err.message);
-            throw Error("Could not retrieve ID");
+
+        await client.login(token);
+        const id = client.user.id;
+        client.destroy().catch(() => {
```
Sorunome: Why not use await / try...catch?

```suggestion
        try {
            await client.destroy();
        } catch (err) {
            log.warn("Failed to destroy client ", id);
        }
```

src/discordas.ts
```diff
@@ -94,6 +94,13 @@ async function run(port: number, fileConfig: DiscordBridgeConfig) {
                 } catch (err) { log.error("Exception thrown while handling \"onAliasQuery\" event", err); }
             },
             onEvent: async (request) => {
+                // tslint:disable-next-line no-any
+                const done = function(resolve: (res: any) => void, reject: (err: Error) => void) {
+                    // tslint:disable-next-line no-invalid-this no-floating-promises
+                    (this as Promise<{}>).then(resolve);
+                    // tslint:disable-next-line no-invalid-this no-floating-promises
+                    (this as Promise<{}>).then(reject);
```
Sorunome: Can you explain this? This looks like we call both the resolve and reject callback, thus resolving and rejecting the promise at the same time?

src/bot.ts
Half-Shot: oopise, yes

src/discordas.ts
Half-Shot: That should be a catch

src/clientfactory.ts
```diff
@@ -64,30 +62,32 @@ export class DiscordClientFactory {
...
-            throw Error("Could not retrieve ID");
+
+        await client.login(token);
+        const id = client.user.id;
+        try {
+            await client.destroy();
+        } catch (err) {
+            log.warn("Failed to destroy client ", id);
+        }
+            log.warn("Failed to destroy client ", id);
```
Sorunome: this looks a tad stray xP

src/discordas.ts
```diff
@@ -95,11 +95,13 @@ async function run(port: number, fileConfig: DiscordBridgeConfig) {
...
             onEvent: async (request) => {
                 // tslint:disable-next-line no-any
+                // This function emulates the Bluebird.done function so the underlying bridge library doesn't
+                // break.
                 const done = function(resolve: (res: any) => void, reject: (err: Error) => void) {
                     // tslint:disable-next-line no-invalid-this no-floating-promises
                     (this as Promise<{}>).then(resolve);
                     // tslint:disable-next-line no-invalid-this no-floating-promises
-                    (this as Promise<{}>).then(reject);
+                    (this as Promise<{}>).catch(reject);
```
Sorunome: so....we run the promise twice now? shoudln't it be `(this as Promise<{}>).then(resolve).catch(reject);`? Soru isn't too sure

Half-Shot: Huh, no I don't think so. This doesn't run the underlying promise, it just provides the callbacks. if we did then().catch then it would also catch if the then() func threw.

PR #450, opened 2019-05-02
Update puppeting 2FA docs.

Half-Shot:


Half-Shot:
Err, not sure how all those commits arrived.

Sorunome:
XD that should work

PR #447, opened 2019-04-29
Worker support

Half-Shot:
Fixes #442, eventually.

PR #454, opened 2019-05-04
Move to matrix-bot-sdk

### LINKED ISSUE #193: Investigate https://github.com/turt2live/matrix-js-bot-sdk as a potential intent replacement
[Intents kinda suck](https://github.com/Half-Shot/matrix-appservice-discord/issues/188) and it would be nice to have the dual benefit of molding them to how discord wants to work, but also have types available to us.

Issues this will face:
* We still need to role our own intent management
* We need to handle registering users
* We need to store user joins more permanently, probably using discord store.

----

On the other hand @Sorunome has done a nice job fixing the join issues in #189, but I can't help but feel this would be a good chance to escape the typing black hole that is matrix-appservice-bridge.

Half-Shot:
Fixes #193 

This moves us from using `matrix-appservice-bridge` to `matrix-bot-sdk`. The benefits are that we are no longer dependent on `matrix-js-sdk`, which should hopefully severely cut down our memory usage. We also gain types for free, as the project is typescript-native. 

However, this will mean that we won't be able to support the old store format for migrations, so I am considering this a 1.0 feature. 

This will be a blocker for #447, as I do not plan to support m-a-b in worker mode.

Things the bot-sdk needs:
- https://github.com/turt2live/matrix-js-bot-sdk/pull/48
- https://github.com/turt2live/matrix-js-bot-sdk/pull/49

- [x] Replace all references of matrix-appservice-bridge with matrix-bot-sdk
- [x] Fix tools
- [x] Fix tests
- [ ] Re-enable third-party support (not supported by the sdk yet)
- [ ] Re-enable support for bridge errors


src/usersyncroniser.ts
```diff
@@ -232,10 +232,10 @@ export class UserSyncroniser {
...
         log.verbose(`State update requested for ${discordUser.id}`);
-        let mxidExtra;
+        let mxidExtra = "";
         if (webhookID) {
             // no need to escape as this mxid is only used to create an intent
-            mxidExtra = "_" + Util.ParseMxid(`@${webhookID}`, false).localpart;
+            mxidExtra = "_"  + webhookID;
         }
         const userState: IUserState = Object.assign({}, DEFAULT_USER_STATE, {
             id: discordUser.id,
```
turt2live: github won't let me comment a line below, but fwiw the bridge has utilities to create these users without tracking all this information. See the `getUserIdForSuffix` function (and `getIntentForSuffix`). Idea is you pass `@discord_.*` to the registration, and for the `ForSuffix` functions you give it `discordUser.id+mxidExtra`

src/bot.ts
```diff
@@ -815,16 +823,16 @@ export class DiscordBot {
...
                 });
-                const afterSend = async (re) => {
-                    this.lastEventIds[room] = re.event_id;
+                const afterSend = async (eventId) => {
+                    this.lastEventIds[room] = eventId;
                     const evt = new DbEvent();
-                    evt.MatrixId = `${re.event_id};${room}`;
+                    evt.MatrixId = `${eventId};${room}`;
                     evt.DiscordId = msg.id;
                     evt.ChannelId = msg.channel.id;
```
Sorunome: It might be worth an idea instead of doing all this to write a join strategy that hooks into the appservice invite one and applies the userstore stuffs correctly

Half-Shot:
@Sorunome I'm quite comfortable with breaking third party support & bridge errors so we can merge this in, and then sprint to fix those afterwards.

src/bot.ts
```diff
@@ -146,10 +138,10 @@ export class DiscordBot {
         if (webhookID) {
             // webhookID and user IDs are the same, they are unique, so no need to prefix _webhook_
             const name = member instanceof Discord.User ? member.username : member.user.username;
-            const nameId = new MatrixUser(`@${name}`).localpart;
-            return this.bridge.getIntentFromLocalpart(`_discord_${webhookID}_${nameId}`);
+            // TODO: We need to sanitize name
+            return this.bridge.getIntentForSuffix(`${webhookID}_${name}`);
```
Sorunome: soru once wrote a santizing routine for the previous thing, unknowingly that it already did it automatically, we could easily add it in again
It is also currently used in mx-puppet-bridge, meaning it is not lost somewhere *yay*

src/db/schema/v8.ts
```diff
@@ -56,33 +53,9 @@ export class Schema implements IDbSchema {
...
-            }
-        }
-        if (migrated !== entrys.length) {
-            log.error(`Didn't migrate all rooms, ${entrys.length - migrated} failed to be migrated.`);
-        } else {
-            log.info("Migrated all rooms successfully");
-        }
+        // XXX: This used to migrate rooms across from the old room store format but
+        // since we moved to the matrix-js-bot-sdk, we can no longer do this. Please
+        // use a 0.X release for this.
```
Sorunome: Is there any way to throw a proper error if old things are present?

src/discordcommandhandler.ts
```diff
@@ -125,9 +125,8 @@ export class DiscordCommandHandler {
             }));
             let errorMsg = "";
             await Promise.all(allChannelMxids.map(async (chanMxid) => {
-                const intent = this.bridge.getIntent();
                 try {
-                    await intent[funcKey](chanMxid, name);
+                    await this.bridge.botIntent.underlyingClient[funcKey + "User"](chanMxid, name);
```
Sorunome: did you test if the `funcKey`s need updating?

src/matrixroomhandler.ts
```diff
@@ -63,22 +53,22 @@ export class MatrixRoomHandler {
...
-        return {
-            getLocation: this.tpGetLocation.bind(this),
-            getProtocol: this.tpGetProtocol.bind(this),
-            getUser: this.tpGetUser.bind(this),
-            parseLocation: this.tpParseLocation.bind(this),
-            parseUser: this.tpParseUser.bind(this),
-            protocols: ["discord"],
-        };
-    }
+    // public get ThirdPartyLookup(): thirdPartyLookup {
```
Sorunome: does the bot sdk not support this yet?

src/matrixroomhandler.ts
```diff
@@ -150,73 +144,72 @@ export class MatrixRoomHandler {
...
-                    },
-                    icon: guild.iconURL || ICON_URL, // TODO: Use icons from our content repo. Potential security risk.
-                    network_id: guild.id,
-                };
-            }),
-            location_fields: ["guild_id", "channel_name"],
-            user_fields: ["username", "discriminator"],
-        };
-    }
+    // public async tpGetProtocol(protocol: string): Promise<thirdPartyProtocolResult> {
```
Sorunome: no support yet, either?

src/presencehandler.ts
```diff
@@ -21,14 +21,13 @@ import { MetricPeg } from "./metrics";
...
-    /* One of: ["online", "offline", "unavailable"] */
-    public Presence: string;
+    public Presence: "online"|"offline"|"unavailable";
     public StatusMsg: string;
     public ShouldDrop: boolean = false;
 }
 
 interface IMatrixPresence {
-    presence?: string;
+    presence?: "online"|"offline"|"unavailable";
```
Sorunome: yay!

src/util.ts
```diff
@@ -423,11 +351,44 @@ export class Util {
         }
         return haveLevel >= requiredLevel;
     }
-}
 
-interface IUploadResult {
-    mxcUrl: string;
-    size: number;
+    public static GetUrlFromMxc(mxc: string, homeserverUrl: string, width: number = 0,
```
Sorunome: doesn't the bot sdk have a function for this? Recalling something like this somehow

Sorunome:
oh, and this MCs

src/db/schema/v8.ts
Half-Shot: How about: 

"Refusing to start bridge, stores are still configured" which means the user has to manually remove the store config file entries in order to start. Hopefully doing that will make them aware enough that they need to migrate things.

src/matrixroomhandler.ts
Half-Shot: Nope.

src/matrixroomhandler.ts
Half-Shot: Nope :(

src/presencehandler.ts
Half-Shot: :)

src/util.ts
Half-Shot: Yeah I remember Travis saying so.

src/db/schema/v8.ts
Sorunome: that sounds like a better solution, yes

src/util.ts
Sorunome: nvm soru just noticed she did this manually, too, so it's probably fine

src/discordcommandhandler.ts
Half-Shot: Good catch!

Half-Shot: Yep, all happy

src/bot.ts
Half-Shot: Hmm, previously we used QP encoding. I'm unsure if the one you linked https://github.com/Sorunome/mx-puppet-bridge/blob/master/src/util.ts#L37 will look very pretty.

src/util.ts
Half-Shot: Funnily enough, we weren't using it at all. I think we *did* switch to the SDKs methods.

Sorunome:
Looks good but lint is failing

src/bot.ts
```diff
@@ -774,6 +771,8 @@ export class DiscordBot {
             const webhook = (await chan.fetchWebhooks())
                             .filterArray((h) => h.name === "_matrix").pop();
             if (webhook && msg.webhookID === webhook.id) {
+                // Filter out our own webhook messages.
+                log.verbose("Not reflecting own webhook messages");
               // Filter out our own webhook messages.
```
auscompgeek: looks like this comment has been duped

src/discordas.ts
```diff
@@ -13,223 +13,187 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
-    reg.setAppServiceToken(AppServiceRegistration.generateToken());
-    reg.setSenderLocalpart("_discord_bot");
-    reg.addRegexPattern("users", "@_discord_.*", true);
-    reg.addRegexPattern("aliases", "#_discord_.*", true);
-    reg.setRateLimited(false);
-    reg.setProtocols(["discord"]);
-    callback(reg);
-}
+function setupLogging() {
+    const logMap = new Map<string, Log>();
```
Sorunome:
```suggestion
    		// log matrix-bot-sdk with our logging
		const logMap = new Map<string, Log>();
		// tslint:disable-next-line:no-any
		const logFunc = (level: string, module: string, args: any[]) => {
			if (!Array.isArray(args)) {
				args = [args];
			}
			// filter out spam logging
			if (args.find((s) => s.includes && s.includes("M_USER_IN_USE"))) {
				return;
			}
			let mod = "bot-ask-" + module;
			const modParts = module.match(/^(\S+)\s(.*)/);
			if (modParts) {
				const MOD_PARTS_NORMAL = 1;
				const MOD_PARTS_EXTRA = 2;
				if (modParts[MOD_PARTS_EXTRA]) {
					args.unshift(modParts[MOD_PARTS_EXTRA]);
				}
				mod = "bot-sdk-" + modParts[MOD_PARTS_NORMAL];
			}
			let logger = logMap.get(mod);
			if (!logger) {
				logger = new Log(mod);
				logMap.set(mod, logger);
			}
			logger[level](...args);
		};
		LogService.setLogger({
			// tslint:disable-next-line:no-any
			debug: (mod: string, args: any[]) => logFunc("debug", mod, args),
			// tslint:disable-next-line:no-any
			info: (mod: string, args: any[]) => logFunc("info", mod, args),
			// tslint:disable-next-line:no-any
			warn: (mod: string, args: any[]) => logFunc("warn", mod, args),
			// tslint:disable-next-line:no-any
			error: (mod: string, args: any[]) => logFunc("error", mod, args),
		});```

Half-Shot: I don't know if we should suggest whole features in reviews..., it's impossible to know what this is replacing.

Sorunome: This is replacing the matrix-bot-sdk logger stuffs you did, it improves upon that slightly (that it uses our logger)

Half-Shot:
(The remaining work will be done in a separate PR, since the goal of this PR is to shift us over)

PR #455, opened 2019-05-04
replace snake_case func names with camelCase in store

Half-Shot:
Part of #377

### DIFF
--- a/src/store.ts
+++ b/src/store.ts
@@ -55,7 +55,7 @@ export class DiscordStore {
         return this.pUserStore;
     }
 
-    public async backup_database(): Promise<void|{}> {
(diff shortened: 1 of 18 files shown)


src/store.ts
```diff
@@ -238,63 +238,6 @@ export class DiscordStore {
             throw err;
         }
     }
-
-    public async get_dm_room(discordId, discordChannel): Promise<string> {
```
Sorunome: meow? huge deleted blocks?

Half-Shot: Yeah we don't use them for anything. Even the DM branch doesn't use them :D

Half-Shot:
(test failed because node ran slow and broke a fragile test)

Issue #336, opened 2018-12-10
Unbridging should remove all ghosts from the room.


Half-Shot:

Sorunome:
implemented

PR #462, opened 2019-05-11
Unbridge fixes

### LINKED ISSUE #283: Unbridge from Discord side
I bridged a matrix room to a discord room using this public bridge: https://t2bot.io/discord/ ...and then left the Matrix room without unbridging.  I'm unable to get back in since it was invite-only, and I was the only one in it, so now it's just the Matrix bot in the room. I am concerned that messages sent in the discord room are still being bridged even though no one is in there.

Is there a way to unbridge from Discord? If not, this would be a valuable feature to have in case this happens.

Half-Shot:
Fixes #283, #336

### DIFF
--- a/src/channelsyncroniser.ts
+++ b/src/channelsyncroniser.ts
@@ -289,7 +289,6 @@ export class ChannelSyncroniser {
         const intent = await this.bridge.getIntent();
         const options = overrideOptions || this.config.channel.deleteOptions;
         const plumbed = entry.remote!.get("plumbed");
-        // tslint:disable-next-line: no-any
 
         await this.roomStore.upsertEntry(entry);
         if (options.ghostsLeave) {
@@ -364,7 +363,6 @@ export class ChannelSyncroniser {
             }
         }
 
-        // Remove entry
         await this.roomStore.removeEntriesByMatrixRoomId(roomId);
     }
 }
--- a/src/channelsyncroniser.ts
+++ b/src/channelsyncroniser.ts
@@ -17,7 +17,7 @@ limitations under the License.
 import * as Discord from "discord.js";
 import { DiscordBot } from "./bot";
 import { Util } from "./util";
-import { DiscordBridgeConfig } from "./config";
+import { DiscordBridgeConfig, DiscordBridgeConfigChannelDeleteOptions } from "./config";
 import { Bridge } from "matrix-appservice-bridge";
 import { Log } from "./log";
 import { DbRoomStore, IRoomStoreEntry } from "./db/roomstore";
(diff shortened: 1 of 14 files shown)


src/channelsyncroniser.ts
```diff
@@ -269,22 +283,24 @@ export class ChannelSyncroniser {
     private async handleChannelDeletionForRoom(
         channel: Discord.TextChannel,
         roomId: string,
-        entry: IRoomStoreEntry): Promise<void> {
+        entry: IRoomStoreEntry,
+        overrideOptions?: DiscordBridgeConfigChannelDeleteOptions): Promise<void> {
         log.info(`Deleting ${channel.id} from ${roomId}.`);
         const intent = await this.bridge.getIntent();
-        const options = this.config.channel.deleteOptions;
+        const options = overrideOptions || this.config.channel.deleteOptions;
```
Sorunome: Why not merge the objects properly? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign

src/channelsyncroniser.ts
```diff
@@ -269,22 +283,24 @@ export class ChannelSyncroniser {
...
         const plumbed = entry.remote!.get("plumbed");
+        // tslint:disable-next-line: no-any
```
Sorunome: stray tslint stuff?

src/provisioner.ts
```diff
@@ -14,35 +14,56 @@ See the License for the specific language governing permissions and
...
+        const remoteRoom = roomsRes[0].remote as RemoteStoreRoom;
+        let roomsToUnbridge: string[] = [];
+        if (rId) {
+            roomsToUnbridge = [rId];
+        } else {
+            // Kill em all.
+            roomsToUnbridge = roomsRes.map((entry) => entry.matrix!.roomId);
+        }
+        await Promise.all(roomsToUnbridge.map( async (roomId) => {
+            return this.channelSync.OnUnbridge(channel, roomId).catch((err) => {
```
Sorunome: why not
```js
try {
    await this.channelSync.OnUnbridge(...);
} catch (err) {
    log.error(err);
}
```

src/util.ts
```diff
@@ -35,7 +35,7 @@ export interface ICommandAction {
     description?: string;
     help?: string;
     params: string[];
-    permission?: PERMISSIONTYPES;
+    permission?: PERMISSIONTYPES | PERMISSIONTYPES[];
```
Sorunome: eeeer, wouldn't it be better to change `type PERMISSIONTYPES = any | any[]; // tslint:disable-line no-any` instead?

src/channelsyncroniser.ts
Half-Shot: The intention of overrideOptions is to literally force a certain config rather than overlay it across.

src/util.ts
Half-Shot: oh yeah, maybe

PR #464, opened 2019-05-11
Lock channels sending a message

Half-Shot:
This is a proposal for an alternative way to avoid message echoback.

### DIFF
--- a/src/bot.ts
+++ b/src/bot.ts
@@ -138,29 +138,27 @@ export class DiscordBot {
         return this.provisioner;
     }
 
-    public lockChannel(channel: Discord.GuildChannel) {
-        const key = `${channel.guild.id}|${channel.id}`;
-        if (this.channelLocks[key]) {
+    public lockChannel(channel: Discord.Channel) {
+        if (this.channelLocks[channel.id]) {
             return;
         }
         let i: NodeJS.Timeout;
         const p = new Promise((resolve) => {
             i = setInterval(resolve, this.config.limits.discordSendDelay);
-            this.channelLocks[key] = {i, p};
+            this.channelLocks[channel.id] = {i, p};
         });
     }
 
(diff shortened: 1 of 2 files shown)


src/bot.ts
```diff
@@ -136,6 +138,34 @@ export class DiscordBot {
...
 
+    public lockChannel(channel: Discord.GuildChannel) {
+        const key = `${channel.guild.id}|${channel.id}`;
+        if (this.channelLocks[key]) {
+            return;
+        }
+        let i: NodeJS.Timeout;
+        const p = new Promise((resolve) => {
+            i = setInterval(resolve, this.config.limits.discordSendDelay);
+            this.channelLocks[key] = {i, p};
```
Sorunome: shouldn't this be outside the promise?

src/bot.ts
```diff
@@ -136,6 +138,34 @@ export class DiscordBot {
         return this.provisioner;
     }
 
+    public lockChannel(channel: Discord.GuildChannel) {
+        const key = `${channel.guild.id}|${channel.id}`;
```
Sorunome: `channel.id` is already a unique key accross all guilds

src/bot.ts
```diff
@@ -200,7 +229,7 @@ export class DiscordBot {
 
         client.on("messageDelete", async (msg: Discord.Message) => {
             try {
-                await Util.DelayedPromise(this.config.limits.discordSendDelay);
+                await this.waitUnlock(msg.channel as Discord.TextChannel);
```
Sorunome: if we just use `channel.id` as key we don't need to cast here, either (which might lead to errors if people DM the discord bot etc.)

src/bot.ts
Half-Shot: Nope, because we need to set the interval pointer *after* it's been set, and `setInterval` needs the resolve function exposed.

Issue #342, opened 2018-12-21
D -> M room mentioning


Sorunome:
If a bridged room already exists with a canonical alias, use that one instead of `#_discord_GUILDID_ROOMID:locahlost`

Sorunome:
implemented!

Issue #397, opened 2019-02-11
Add option to disable #_discord_ bridging


Sorunome:
Some poeple don't want that

PR #467, opened 2019-05-12
D->M room mentions

Sorunome:
This PR does two things (D->M parsing):

1. fix an unreported bug where emoji insertions in plain body weren't replaced correctly
2. use the canonical alias for channels, if available #342

### DIFF
--- a/src/discordmessageprocessor.ts
+++ b/src/discordmessageprocessor.ts
@@ -39,7 +39,7 @@ const CHANNEL_INSERT_REGEX = /\x01chan\x01([0-9]*)\x01/;
 const ID_CHANNEL_INSERT_REGEX = 1;
 
 export class DiscordMessageProcessorOpts {
-    constructor(readonly domain: string, readonly bot?: DiscordBot, readonly bridge?: Bridge) {
+    constructor(readonly domain: string, readonly bot?: DiscordBot) {
 
     }
 }
@@ -64,7 +64,7 @@ export class DiscordMessageProcessor {
     constructor(opts: DiscordMessageProcessorOpts, bot: DiscordBot | null = null) {
         // Backwards compat
         if (bot !== null) {
-            this.opts = new DiscordMessageProcessorOpts(opts.domain, bot, opts.bridge);
+            this.opts = new DiscordMessageProcessorOpts(opts.domain, bot);
         } else {
             this.opts = opts;
         }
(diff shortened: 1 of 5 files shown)


src/discordmessageprocessor.ts
```diff
@@ -270,7 +271,8 @@ export class DiscordMessageProcessor {
             try {
                 const mxcUrl = await this.opts.bot!.GetEmoji(name, animated, id);
                 if (html) {
-                    replace = `<img alt="${name}" title="${name}" height="${EMOJI_SIZE}" src="${mxcUrl}" />`;
+                    replace = `<img alt="${escapeHtml(name)}" title="${escapeHtml(name)}" ` +
```
Half-Shot: You've escaped name 3 times, could we make this a `const` above?

src/discordmessageprocessor.ts
```diff
@@ -279,18 +281,48 @@ export class DiscordMessageProcessor {
...
+            const id = results[ID_CHANNEL_INSERT_REGEX];
+            let replace = "";
+            const channel = msg.guild.channels.get(id);
+            if (channel) {
+                const name = "#" + channel.name;
+                let alias = `#_discord_${msg.guild.id}_${id}:${this.opts.domain}`;
+                try {
+                    // we don't bother here much of checking if objects exists, as the error is catched
+                    // and we use the default alias then, as we wanted to anyways
+                    const matrixRoom = (await this.opts.bot!.ChannelSyncroniser.GetRoomIdsFromChannel(channel))[0];
```
Half-Shot: If this first roomid is to fail, do we retry with other roomids?

test/test_discordmessageprocessor.ts
```diff
@@ -404,7 +399,18 @@ describe("DiscordMessageProcessor", () => {
                 name: "blah",
             };
             const reply = processor.InsertEmoji(content);
-            Chai.assert.equal(reply, "\x01blah\x011\x011234\x01");
+            Chai.assert.equal(reply, "\x01emoji\x01blah\x011\x011234\x01");
+        });
+    });
+    describe("InsertChannel", () => {
+        it("inserts channels to their post-parse flag", () => {
```
Half-Shot: do we want to test multiple channel mentions to ensure it doesn't get itself confused?

src/discordmessageprocessor.ts
Sorunome: nope, should that be added?

test/test_discordmessageprocessor.ts
Sorunome: good idea

src/discordmessageprocessor.ts
Sorunome: Also, this should move into channel sync or something anyways as it could be needed for #431 and stuffs

Sorunome: Yeah, soru will re-write some things here also prepping for #397

Half-Shot: kk

PR #471, opened 2019-05-16
Replace istanbul with nyc

### LINKED ISSUE #461: Replace istanbul with nyc
istanbul is unmaintained

Half-Shot:
Fixes #461


Sorunome:
The package-lock.json is suddenly way huger, what's up with that? sudden additional dependencies? or are there now unused packages in there?

Half-Shot:
New packages which apparently depend on a shit ton of things, let me move some of the packages into devDependencies

PR #472, opened 2019-05-16
Ban guilds and channels from being bridged

### LINKED ISSUE #362: Add a way to block a guild (and optional channel) from being bridged
Something like `!block <guild_id> [channel_id]` from a bridge admin, accompanied by an `!unblock` command.

Half-Shot:
Fixes #362


Sorunome:
This should also be added in #467's chanenlsync GetAliasFromChannel to return null, if it is banned from being bridged

src/bot.ts
```diff
@@ -227,6 +227,9 @@ export class DiscordBot {
 
         client.on("messageDelete", async (msg: Discord.Message) => {
             try {
+                if (await this.store.isGuildChannelBanned(msg.guild.id, msg.channel.id)) {
```
Sorunome: go only over channel id, `msg.guild` may even be undefined.

the channel ID is already a 100% unique way to identify all channels no matter what

Half-Shot: So the reason we use guild id in the db is for readability as much as anything. It's important to preserve the guild => channel mapping for that purpose. I guess we could search by either channel or guild tho.

Sorunome: this will severely break the bot once we add something for non-guild channels down the road, though (e.g. command interface via DM?) due to those not having a guild.

Half-Shot:

PR #473, opened 2019-05-16
Add indexes to tables

### LINKED ISSUE #382: Create indexes in postgres
Currently there's no indexes, which means a lot of sequential scans. Here's the indexes I created:

```sql
create index _t2b_emoji_id on emoji(emoji_id);
create index _t2b_emoji_mxc on emoji(mxc_url);
create index _t2b_event_store_mxid on event_store(matrix_id);
create index _t2b_event_store_discordid on event_store(discord_id);
create index _t2b_discord_msg_store_msgid on discord_msg_store(msg_id);
```

and table row counts at the time:

|table|rows|
|-----|------|
|`emoji`|4314|
|`event_store`|8701140|
|`discord_msg_store`|7618470|
|`user_id_discord_id`|0|
|`discord_id_token`|0|
|`dm_rooms`|0|

Half-Shot:
Fixes #382

### DIFF
--- a/src/db/schema/v10.ts
+++ b/src/db/schema/v10.ts
@@ -1,5 +1,5 @@
 /*
-Copyright 2018 matrix-appservice-discord
+Copyright 2019 matrix-appservice-discord
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
--- a/src/db/schema/v10.ts
+++ b/src/db/schema/v10.ts
@@ -1,0 +1,54 @@
+/*
+Copyright 2018 matrix-appservice-discord
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import {IDbSchema} from "./dbschema";
+import {DiscordStore} from "../../store";
+import { Log } from "../../log";
+
+const log = new Log("SchemaV10");
+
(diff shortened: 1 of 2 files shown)


src/db/schema/v10.ts
```diff
@@ -0,0 +1,54 @@
+/*
+Copyright 2018 matrix-appservice-discord
```
Sorunome: it's 2019 xP

src/db/schema/v10.ts
```diff
@@ -0,0 +1,54 @@
...
+        idx_room_entries_id: ["room_entries", "id"],
+        idx_room_entries_matrix_id: ["room_entries", "matrix_id"],
+        idx_room_entries_remote_id: ["room_entries", "remote_id"],
+    };
+
+    public async run(store: DiscordStore): Promise<void> {
+        try {
+            await Promise.all(Object.keys(this.INDEXES).map(async (indexId: string) => {
+                const ids = this.INDEXES[indexId];
+                return store.db.Exec(`CREATE INDEX ${indexId} ON ${ids[0]}(${ids[1]})`);
```
Sorunome: what if we have multiple combinations to index? we don't have that, do we?

src/store.ts
```diff
@@ -29,7 +27,9 @@ import { DbUserStore } from "./db/userstore";
 import {
     RoomStore, UserStore,
 } from "matrix-appservice-bridge";
+
 const log = new Log("DiscordStore");
+export const CURRENT_SCHEMA = 10;
```
Sorunome: why the location switch?

src/db/schema/v10.ts
Half-Shot: BAH HUMBUG

src/db/schema/v10.ts
Half-Shot: Nope. If we ever do, then we can just improve upon this in the next schema

src/store.ts
Half-Shot: I was annoyed that we had the schema in the middle of our imports

Half-Shot:
>did you test with both sqlite and postgres?

The syntax is the same but...no I didn't. I'll do that

PR #474, opened 2019-05-16
Fix locks

Half-Shot:


src/bot.ts
```diff
@@ -142,17 +142,22 @@ export class DiscordBot {
         if (this.channelLocks[channel.id]) {
             return;
         }
-        let i: NodeJS.Timeout;
+
         const p = new Promise((resolve) => {
-            i = setInterval(resolve, this.config.limits.discordSendDelay);
-            this.channelLocks[channel.id] = {i, p};
+            if (!this.channelLocks[channel.id]) {
+                return;
```
Sorunome: resolve(); before return, just in case?

src/bot.ts
```diff
@@ -142,17 +142,22 @@ export class DiscordBot {
...
+            const i = setInterval(resolve, this.config.limits.discordSendDelay);
+            this.channelLocks[channel.id].i = i;
         });
+
+        this.channelLocks[channel.id] = {i: null, p};
     }
 
     public unlockChannel(channel: Discord.Channel) {
         const lock = this.channelLocks[channel.id];
         if (lock) {
```
Sorunome: `if (lock && lock.i)`, that way the `!` in `lock.i!` is unneeded in next line

src/bot.ts
Half-Shot: The return is designed to save us setting up a timer if we have already unlocked the channel before the promise is processed. It's astronomically unlikely, but we don't need to setup a timer if it's been deleted.

Sorunome: soru meany
```js
if (!this.channelLocks[channel.id]) {
    resolve();
    return;
}
```
that the promise still resolves if we don't need a timer

Half-Shot: ooooh, yes

PR #480, opened 2019-05-22
Metrics

Half-Shot:
This PR enables metrics support for the bridge. The way we are doing this is slightly weird. Rather than passing around a metrics instance to all of the handlers, I've just made a MetricPeg static class, which will host a `IBridgeMetrics` instance of your choosing. Personally, this feels more elegant.


src/clientfactory.ts
```diff
@@ -110,4 +108,12 @@ export class DiscordClientFactory {
             return this.botClient;
         }
     }
+
+    public bindMetricsToChannel(channel: TextChannel) {
+        // tslint:disable-next-line:only-arrow-functions
+        channel.send = function() {
+            MetricPeg.get.remoteCall("channel.send");
+            return channel.send.apply(channel, arguments);
```
Sorunome: doesn't this recurse? What if we call `bindMetricsToChannel` twice on a channel object, won't it do the metrics twice?

src/bot.ts
```diff
@@ -389,6 +392,7 @@ export class DiscordBot {
             }
             const channel = guild.channels.get(room);
             if (channel && channel.type === "text") {
+                this.ClientFactory.bindMetricsToChannel(channel as Discord.TextChannel);
```
Sorunome: shouldn't we  always bind the channel on message processing? Or even better, right when the message enters via the event listeners?

src/bot.ts
```diff
@@ -766,14 +776,13 @@ export class DiscordBot {
         let rooms;
         try {
             rooms = await this.channelSync.GetRoomIdsFromChannel(msg.channel);
+            if (rooms === null) { throw Error(); }
```
Sorunome: new lines for if-condition due to `{}`?

src/metrics.ts
```diff
@@ -0,0 +1,140 @@
...
+export class PrometheusBridgeMetrics implements IBridgeMetrics {
+    private metrics;
+    private remoteCallCounter: Counter;
+    private storeCallCounter: Counter;
+    private presenceGauge: Gauge;
+    private remoteRequest: Histogram;
+    private matrixRequest: Histogram;
+    private requestsInFlight: Map<string, number>;
+    private bridgeGauges: IBridgeGauges = {
+        matrixGhosts: 0,
```
Sorunome: are these autofilled somehow?

src/bot.ts
Half-Shot: Nope, because we only call the *send* function when sending stuff that came via matrix. Though I guess we should also bind in the event listeners too.

src/clientfactory.ts
Half-Shot: Yes hello I'm dumb, I wonder how we do this sanely.

Half-Shot:
>Soru is a bit confused, are these metrics only in-memory? are they saved somehow? if yes, how? If it is just a json blob file thingy like old user and room store, wouldn't that introduce a significant bottleneck?

By default it blackholes the metrics unless you enable prometheus in the config. If enabled, the bridge will simply add another endpoint onto the bridge's http listener which will spit out a line seperated list of values which are kept in memory. A prometheus instance will then poll this endpoint every X seconds. It's not at all memory intensive since reads are done asyncronously and writes are just updating integer values.

src/metrics.ts
Half-Shot: Yes, by the bridge lib. Though some of them will remain 0

src/bot.ts
Half-Shot: I think I was being lazy, sorry.

Sorunome:
> > Soru is a bit confused, are these metrics only in-memory? are they saved somehow? if yes, how? If it is just a json blob file thingy like old user and room store, wouldn't that introduce a significant bottleneck?
> 
> By default it blackholes the metrics unless you enable prometheus in the config. If enabled, the bridge will simply add another endpoint onto the bridge's http listener which will spit out a line seperated list of values which are kept in memory. A prometheus instance will then poll this endpoint every X seconds. It's not at all memory intensive since reads are done asyncronously and writes are just updating integer values.

ah, so you need to run another program (prometheus server) to read out the metrics? makes sense, then

Sorunome:
why don't we bind `this.clientFactory.bindMetricsToChannel(msg.channel as Discord.TextChannel);` at the start of all those discord event listeners? (like edit, delete etc)

Half-Shot:
# :boom:

PR #481, opened 2019-05-22
Profile Caching

Half-Shot:
Requires #480 because I based the branch off that.

This makes the change so that we cache profile information about users and don't do a lookup per request. It also finally introduces a timed cache type.


src/structures/timedcache.ts
```diff
@@ -0,0 +1,103 @@
...
+
+    public clear(): void {
+        this.map.clear();
+    }
+
+    public delete(key: K): boolean {
+        return this.map.delete(key);
+    }
+
+    public forEach(callbackfn: (value: V, key: K, map: Map<K, V>) => void): void {
```
Sorunome: why not make this async, in case the callback is async?

src/structures/timedcache.ts
```diff
@@ -0,0 +1,103 @@
...
+        for (const item of this) {
+            callbackfn(item[1], item[0], this);
+        }
+    }
+
+    public get(key: K): V | undefined {
+        const v = this.map.get(key);
+        if (v) {
```
Sorunome: everywhere else we check for `!== undefined`, why not here? This would cause things to break in case of `null` or `false`

src/structures/timedcache.ts
```diff
@@ -0,0 +1,103 @@
...
+                let filteredValue: V|undefined;
+                // Loop if => 1) no item/filteredvalue OR item && !filteredValue && not done
+                while (!item) {
+                    item = iterator.next();
+                    if (item.done) {
+                        break;
+                    }
+                    filteredValue = this.filterV(item.value[1]);
+                }
+                if (!item || filteredValue === undefined) {
```
Sorunome: won't this prematurely stop iterating over the cache? Shouldn't we instead `while (!item && filteredValue === undefined)` ?

src/matrixeventprocessor.ts
```diff
@@ -176,7 +178,10 @@ export class MatrixEventProcessor {
         }
 
         await this.discord.send(embedSet, opts, roomLookup, event);
-        await this.sendReadReceipt(event);
+        // Don't await this.
```
Sorunome: yay for these comments, makes it way easier to read along! :+1:

src/matrixeventprocessor.ts
```diff
@@ -367,6 +360,36 @@ export class MatrixEventProcessor {
...
+        const intent = this.bridge.getIntent();
+        let profile: {displayname: string, avatar_url: string} | undefined;
+        try {
+            profile = this.mxUserProfileCache.get(`${roomId}:${userId}`);
+            if (!profile) {
+                log.verbose(`Profile ${userId}:${roomId} not cached`);
+                profile = await mxClient.getStateEvent(roomId, "m.room.member", userId);
+            }
+            if (profile) {
+                this.mxUserProfileCache.set(`${roomId}:${userId}`, profile);
```
Sorunome: move cache setting into `if(!profile) {...}` and that other else condition then another `if(!profile){...}`? Else our timed cache doesn't matter as we keep on resetting the cache timeout

src/matrixeventprocessor.ts
```diff
@@ -418,13 +442,6 @@ export class MatrixEventProcessor {
             }
             // Let it fall through.
         }
-        if (!profile) {
```
Sorunome: we assume here now that the `profile` exists, but it can still not exist due to the code up there? why remove this?

src/matrixeventprocessor.ts
```diff
@@ -59,12 +62,14 @@ export class MatrixEventProcessor {
...
     private matrixMsgProcessor: MatrixMessageProcessor;
     private mxCommandHandler: MatrixCommandHandler;
+    private mxUserProfileCache: TimedCache<string, {displayname: string, avatar_url: string}>;
 
     constructor(opts: MatrixEventProcessorOpts, cm?: MatrixCommandHandler) {
         this.config = opts.config;
         this.bridge = opts.bridge;
         this.discord = opts.discord;
         this.matrixMsgProcessor = new MatrixMessageProcessor(this.discord);
+        this.mxUserProfileCache = new TimedCache(PROFILE_CACHE_LIFETIME);
```
Sorunome: Okay, we cache the profiles now, great! Only problem.....don't we need to flush the cache for that user on `m.room.member` changes? This doesn't seem to happen anywhere?

Sorunome: and/or just update to the new state we already have then

src/structures/timedcache.ts
Half-Shot: I'm following the `Map` interface which doesn't let me :(. Though I *guess* I could check to see if callbackfn return a promise and awaiting on that.

src/structures/timedcache.ts
Half-Shot: Good catch!

src/structures/timedcache.ts
Half-Shot: On second hand, a forEach shouldn't be handling promises. If you want to use promises, use a `for (let a of b) {} ` block

src/matrixeventprocessor.ts
Half-Shot: There isn't any need to attempt to fetch the profile twice? If we failed the first time, we just use the mxid and continue.

src/structures/timedcache.ts
Sorunome: Right, true!

src/structures/timedcache.ts
Sorunome: Now this if-condition won't ever run, you could just remove it.

If it breaks due to `item.done` being true, then `if(!item)` wouldn't run, and with filteredValue it can't be undefined at the end of this

src/matrixeventprocessor.ts
```diff
@@ -205,6 +210,11 @@ export class MatrixEventProcessor {
             msg += `set the topic to \`${event.content!.topic}\``;
         } else if (event.type === "m.room.member") {
             const membership = event.content!.membership;
+            if (membership === "join") {
+                this.mxUserProfileCache.delete(`${event.room_id}:${event.sender}`);
```
Sorunome: Alternatively, we could also set the new `membership` right away? (not 100% sure if it would need slight re-formatting)

Half-Shot: Huh? Not really, this is only used for setting webhook displaynames iirc so probably not much we can do?

Sorunome: eeer, soru means `event.content`, sorry!

Half-Shot: OH

Half-Shot: i am dumb

PR #482, opened 2019-05-30
Oauth identify

Half-Shot:
(I though we had a issue for this but it can't be found).

This allows matrix users to "sign in" as Discord users, so that matrix users and discord users can be mapped together which should allow for more integration features in the future. This is a separate feature to puppeting.


src/db/dbaccesstoken.ts
```diff
@@ -0,0 +1,93 @@
...
+        this.DiscordId = item.discord_id;
+        this.AccessToken = item.access_token;
+        this.RefreshToken = item.refresh_token;
+        this.ExpiresIn = item.expires_in;
+        this.CreatedAt = item.created_at;
+        return true;
+    }
+
+    public async Update(store: DiscordStore): Promise<void> {
+        throw new Error("Delete is not implemented");
```
Sorunome: Shouldn't it say "Update is not implemented"?

src/db/dbaccesstoken.ts
```diff
@@ -0,0 +1,93 @@
...
+        } else {
+            throw new Error("Must specify either matrix_id or discord_id");
+        }
+        const rows = await store.db.All(query, {
+            id,
+        });
+        this.Result = rows.length > 0;
+    }
+
+    public async Insert(store: DiscordStore): Promise<void> {
```
Sorunome: shouldn't insert auto-check if it already exists, and if so, only update? To prevent multiple entries. Or are they wanted?

src/discordas.ts
```diff
@@ -180,7 +182,10 @@ async function run(port: number, fileConfig: DiscordBridgeConfig) {
 
     const discordbot = new DiscordBot(botUserId, config, bridge, store);
     const roomhandler = discordbot.RoomHandler;
-    const eventProcessor = discordbot.MxEventProcessor;
+    const eventProcessor = new MatrixEventProcessor(
+        new MatrixEventProcessorOpts(config, bridge, discordbot),
+        new MatrixCommandHandler(discordbot, bridge, config, store),
```
Sorunome: why can't we have the event processor create the command handler?

src/matrixcommandhandler.ts
```diff
@@ -143,6 +151,21 @@ export class MatrixCommandHandler {
             },
         };
 
+        if (this.oauthHandler) {
+            // TODO: Only allow for PM rooms.
+            actions.identify = {
+                description: "Identify your discord identity with the bridge",
+                params: [],
+                run: async () => {
+                    if (!this.oauthHandler) {
```
Sorunome: isn't this check kinda redundant?

src/matrixcommandhandler.ts
```diff
@@ -143,6 +151,21 @@ export class MatrixCommandHandler {
             },
         };
 
+        if (this.oauthHandler) {
+            // TODO: Only allow for PM rooms.
```
Sorunome:
```ts
if (context.rooms.remote) {
    // deny this command
}
```
could be inside the actual run to give a meaningful message

src/oauthhandler.ts
```diff
@@ -0,0 +1,126 @@
...
+export class OAuthHandler {
+    private stateUserMap: Map<string, string>;
+    constructor(private config: DiscordBridgeConfigAuth, private bridge: Bridge, private store: DiscordStore) {
+        this.stateUserMap = new Map();
+    }
+
+    public bindEndpoint() {
+        this.bridge.addAppServicePath({
+            handler: (req: Request, res: Response) => {
+                this.onGetOAuth(req, res).catch((ex) => {
```
Sorunome: why `promise.catch(...)` instead of

```ts
try {
    await this.onGetOAuth(req, res);
} catch (ex) {
    // handle exception here
}
```

src/oauthhandler.ts
```diff
@@ -0,0 +1,126 @@
...
+        return ident.id;
+    }
+
+    private scopes(escape: boolean = false) {
+        return ["identify", "guilds", "guilds.join"].join(escape ? "%20" : " ");
+    }
+
+    private get authUrl() {
+        return `${URL_AUTH}?client_id=${this.config.clientID}&response_type=code` +
+        `&redirect_uri=${this.config.oAuthUrl}&scope=${this.scopes(true)}`;
```
Sorunome: scope doesn't need to be that integer bitfield? It can be easily calculated using discord.js

src/oauthhandler.ts
```diff
@@ -0,0 +1,126 @@
...
+            const atRes = JSON.parse(await (request.post({
+                formData,
+                simple: true,
+                url: URL_TOKEN,
+            }))) as IDiscordTokenResponse;
+            const dataObj = new DbAccessToken();
+            dataObj.DiscordId = await this.getIdentity(atRes.access_token);
+            dataObj.AccessToken = atRes.access_token;
+            dataObj.RefreshToken = atRes.refresh_token;
+            dataObj.ExpiresIn = atRes.expires_in || 1;
```
Sorunome: what do we do when the token expires?

Half-Shot: Nothing. I guess we could invalidate the mapping but that seems kinda rude and I don't think it's a good idea to ask users to reauth every few weeks to keep their mappings. Perhaps we could just require reauth for any *new* explicit actions (if I can think of any) and anything automatic like power levels will continue to happen.

Half-Shot:
Some of this PR seems to have been mangled with old-develop

Half-Shot:

PR #484, opened 2019-06-04
remove old config values

Sorunome:


Half-Shot:
#454 already does this, and also removes the migration code.

Sorunome:
@Half-Shot should we merge that one soon, then? what about metrics?

Half-Shot:
Well, we still need these values for migrating 0.4 -> 0.5.

0.5.1 will only contain metrics stuff, but otherwise should be treated as 0.5

0.6 will contain the bot-sdk stuff hopefully, and will probably be called 1.0

Half-Shot:
>#454 already does this, and also removes the migration code.

apparently this was a lie, merging.

PR #485, opened 2019-06-04
Override config with environment variables

pacien:
This adds the possibility for the user to override configuration values by setting environment variables.

For example, one can define the value for `auth.botToken` by setting `APPSERVICE_DISCORD_AUTH_BOT_TOKEN=iDislikeTypescriptAndThisNodejsMadness` in the environment.

A unit test has been added for the proposed changes.
This PR also fixes some invalid calls in some other tests.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -14,6 +14,10 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+const ENV_PREFIX = "APPSERVICE_DISCORD";
+const ENV_KEY_SEPARATOR = "_";
+const ENV_VAL_SEPARATOR = ",";
+
 /** Type annotations for config/config.schema.yaml */
 export class DiscordBridgeConfig {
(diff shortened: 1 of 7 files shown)


src/config.ts
```diff
@@ -27,18 +31,46 @@ export class DiscordBridgeConfig {
...
+                this.applyConfig(newConfig[key], configLayer[key]);
+            } else {
+                configLayer[key] = newConfig[key];
+            }
+        });
+    }
+
+    /**
+     * Override configuration keys defined in the supplied environment dictionary.
+     * @param environment environment variable dictionary
```
Half-Shot: What's path and configLayer

pacien: Doc added.

pacien:
@Half-Shot wrote:
> Broadly looks good as is for me, but can we get a README note to say that this feature exists?

Sure, added.

PR #491, opened 2019-06-07
Matrix event types

V02460:
The Discord bridge uses Matrix events throughout its code base, so for a robust bridge it is important to get the handling of them right. Currently there is only one type `IMatrixEvent` for every possible Matrix event. This leads to many places where an attribute is just assumed to be there (!) and to code where it is not clear which type of event can be handled. I added tighter types around Matrix events so the benefits of TypeScript can be reaped.

A second issue is that the events are not sanitized and error checking is scattered all over the code (which is not good^^). I would propose to add a runtime type checker for this purpose and I found [runtypes](https://github.com/pelotom/runtypes) to be pretty promising. Please let me know if you agree!

This PR refactors the production code to use coarser Matrix event types. I did not adapt the test sources and did not made it use a runtime type checker for I want to wait for feedback before doing the work.

**I don't expect this code to be runnable and error-free at this point.** I started this PR to give you a first impression and have a concrete discussion about it. If everyone is happy with it I will start to make it pretty.


V02460:
(Sry, had no time for rebase, but still wanted to get it out)

Half-Shot:
Thank you for taking the time to do this, I think we've majorly missed a trick by not doing this from the get-go. I think it's definitely a good idea though I'll ask @Sorunome to also validate it.

As far as `runtypes` is concerned, it's something that sounds very useful and I'll do my research on it with @Sorunome to decide if it works for us.

I do kind of wish this wasn't restricted to the discord bridge, and instead we had a general 'matrix-events' library that had type definitions and validation built in.

Sorunome:
Very good idea, yes! Will review later, just.....how much will the bot-sdk handle this stuff already?

Half-Shot:
> Very good idea, yes! Will review later, just.....how much will the bot-sdk handle this stuff already?

It doesn't. I don't know if @turt2live has *plans* though.

Sorunome:
This def. sounds like the thign bot-sdk should handle :thinking:

Issue #478, opened 2019-05-21
D->M format spoilers correctly


Sorunome:
use font-tag with black fore-and-background

Sorunome:
Implemented in #502

PR #502, opened 2019-06-17
handle spoilers

Sorunome:
Handle spoilers as per [MSC2010](https://github.com/matrix-org/matrix-doc/pull/2010)

Implements #478


Half-Shot:
Oh um before you merge, could you add a little note somewhere in the code that the MSC is still pending

PR #506, opened 2019-06-23
Fix room visibility and add other privacy options

### LINKED ISSUE #396: room.defaultVisibility setting not working
if setting room.defaultVisibility is set to `private` it actually doesn't work

Half-Shot:
Fixes #396

### DIFF
--- a/src/matrixroomhandler.ts
+++ b/src/matrixroomhandler.ts
@@ -107,7 +107,7 @@ export class MatrixRoomHandler {
         this.bridge.getIntent().getClient().setRoomDirectoryVisibilityAppService(
             channel.guild.id,
             roomId,
-            "public",
+            this.config.room.defaultVisibility,
         );
         await this.discord.ChannelSyncroniser.OnUpdate(channel);
         const promiseList: Promise<void>[] = [];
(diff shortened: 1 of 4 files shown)


src/matrixroomhandler.ts
```diff
@@ -266,11 +266,19 @@ export class MatrixRoomHandler {
             initial_state: [
                 {
                     content: {
-                        join_rule: "public",
+                        // TODO: We only really support "public", since we don't know who to invite.
```
Sorunome: this is rather.....meh

Half-Shot: Not much we can do:
 - We don't know who is trying to join
 - We can't yet map matrix users to discord users to determine who is allowed. We also can't "whitelist users".

Sorunome: It might be better to not have that option at all, then?

Sorunome: that or a script for the AS bot to invite someone in conjunction with this

Half-Shot: >invite someone

the problem is who :). But I agree, we could probably remove this option since it's useless.

Sorunome: How is it a problem? Something like adminme but instead it invites someone, shouldn't be far-fetched

Half-Shot: Seems like more of a job for provisioning than aliases though.

Sorunome: what do you mean? An `npm run inviteme` script would fix the issue for now, would it not?

Half-Shot: So what I would do in this instance is write a script that provisions an invite only room and invites the user who invoked it. Configuring your bridge to do invite only rooms on alias requests is weird because it's always going to access deny you.

The *other* option is configuring the bridge to always invite the same user(s) when creating these rooms, which might work for self hosted instances.

Sorunome: > invites the user who invoked it

How do we know who revoked an alias query?

Half-Shot: We don't know who *in*voked the alias :(

Half-Shot:

PR #512, opened 2019-06-26
Reliable discord

V02460:
Adapts the Discord bridge to the changed error handling of the Bridge SDK to support signaling of permanent errors.

The SDK slightly changes the semantics of rejected request promises on the event callback. Previously a rejected promise would only be used for event queuing by the bridge and not trigger any further action. Now a special case is introduced for the rejection where this is no longer the case. The three possible outcomes are now as follows:

**Fulfilled**: The event was successfully handled by the bridge. No further action is required.
**Rejected with `EventNotHandledError`** ¹: The bridge decided that the event can never ever be handled and instructs the SDK to signal this permanent error to the room.
**Otherwise rejected**: The bridge encountered a failure while processing the event. The bridge decided to silently swallow the event. *This is the legacy behavior*.

The interface change itself is mostly downwards compatible, so old behavior still present in the source of the bridge will behave as previously. I was pushing some things from A to B here, though.
****************
¹ Including `EventNotHandledError`'s subtypes


Sorunome:
aaaaa without looking much at the code yet, we purposefully got away from chai-as-promised

for tests, you can use instead try...catch using await!

```ts
try {
    await someTestFunc();
    throw "should not error";
} catch (err) {
    expect(err).to.not.equal("should not error");
}
```
for example

Sorunome:
@Half-Shot @V02460 how would this work with matrix-bot-sdk instead of matrix-appservice-bridge?

src/discordas.ts
```diff
@@ -55,8 +68,19 @@ function generateRegistration(reg, callback)  {
     callback(reg);
 }
 
-// tslint:disable-next-line no-any
-type callbackFn = (...args: any[]) => Promise<any>;
+interface IBridgeCallbacks {
+    onAliasQueried?: (alias: string, roomId: string) => Promise<void>;
```
Sorunome: why have this and the next one as optional if they are set always anyways? That results in unneeded usage of `!` all over

src/discordas.ts
```diff
@@ -209,3 +224,70 @@ async function run(port: number, fileConfig: DiscordBridgeConfig) {
         process.exit(1);
     }
 }
+
+/**
+ * Logs an error which occured during event processing.
+ *
+ * Depending on the error type different log levels are hardcoded.
+ */
+function logOnEventError(err: Error): void {
```
Sorunome: wouldn't it be cleaner to make this a private method? Same goes for the other defined methods below

src/discordas.ts
```diff
@@ -209,3 +224,70 @@ async function run(port: number, fileConfig: DiscordBridgeConfig) {
...
+        )
+    ;
+}
+
+/**
+ * Builds a custom BridgeContext with rooms known to the bridge.
+ */
+async function buildOwnContext(
+    roomId: string,
+    store: DiscordStore,
```
Sorunome: if `buildOwnContext` is a private method you don't have to pass over store, right?

src/util.ts
```diff
@@ -421,3 +421,35 @@ interface IUploadResult {
...
     size: number;
 }
+
+// Type type
+type Type = Function;  // tslint:disable-line ban-types
+
+/**
+ * Returns true if `obj` is subtype of at least one of the given types.
+ */
+export function instanceofsome(obj: object, types: Type[]): boolean {
```
Sorunome: can't you make these public static functions of the Util class like all the other ones? Then we stay consistent with `Util.wrap()` and `Util.instanceofsome()`

src/discordas.ts
```diff
@@ -78,53 +102,42 @@ async function run(port: number, fileConfig: DiscordBridgeConfig) {
...
+                    logOnEventError(err);
+                    request.reject(err);
+                } finally {
+                    recordRequestOutcome(request);
                 }
             },
-            onLog: (line, isError) => {
-                log.verbose("matrix-appservice-bridge", line);
+            onLog: (text: string, isError: boolean): void => {
+                const logText = (isError ? log.error : log.verbose).bind(log);
```
Half-Shot: This is quite confusing to read, I'd just use an if statement.

src/matrixeventprocessor.ts
```diff
@@ -123,21 +142,23 @@ export class MatrixEventProcessor {
                     await this.ProcessMsgEvent(event, srvChanPair[0], srvChanPair[1]);
                     return;
                 } catch (err) {
-                    log.warn("There was an error sending a matrix event", err);
-                    return;
+                    throw wrap(err, Error, "There was an error sending a matrix event");
```
Half-Shot: `wrap` is not nearly clear enough as a name. What are you wrapping?

V02460:
> Where does the resending happen? within matrix-appservice-sdk?
No, it's what the home server is doing. If I got it right, it caches  all messages which are not handled by the AS and resends them after like 20 min.

> how would this work with matrix-bot-sdk instead of matrix-appservice-bridge? As in, could those imported errors be easily re-implemented?
A cautious yes to that question. I didn't take a look at matrix-bot-sdk yet. There must be some mechanism to notify the HS, so the basics should be there.

src/matrixeventprocessor.ts
V02460: I kind of like the name. It is wrapping the original error into a new one (while preserving the stacktrace).

V02460: A low-level `IOError` at a higher level is more likely a `RequestError` or `UserRegistrationError`. That's why it is nice to use this functionality: It gives more meaning to an error.

src/discordas.ts
V02460: The bind makes it all bad :smile:

src/util.ts
V02460: Oh indeed, I didn't conform to consistency there.

Why is this Util class there at all btw? These functions have their own util module, so they are already wrapped nicely. This does seem to only add noise to the code using this methods, or is there some JS quirk which makes it necessary?

src/discordas.ts
V02460: It's not a method, so I have to pass the store (which is scoped to the `run` function).

src/discordas.ts
V02460: If I get TS right, this function is private to the module (file) and is not in the global namespace, so everything should be fine with it.

src/discordas.ts
V02460: No need for that, indeed. I mixed it up with the actual callbacks of the bridge, which are optional.

V02460:
> we aren't handling bot.ts line 477, where failures are most likely to occur since we will just log and drop any messages that failed to send remotely.

Good, that you tell me. I fear there is no nice way to be sure to catch all error cases reliably (but exception bubbling in general helps a lot here). I will look into it.

src/util.ts
```diff
@@ -421,3 +421,56 @@ interface IUploadResult {
...
```
Half-Shot: This should be `isInstanceOfTypes`, personally

src/util.ts
```diff
@@ -421,3 +421,56 @@ interface IUploadResult {
...
+    return types.some((type) => obj instanceof type);
+}
+
+/**
+ * Append the old error message to the new one and keep its stack trace.
+ * Example:
+ *     throw wrap(e, HighLevelError, "This error is more specific");
+ */
+export function wrap<T extends Error>(
```
Half-Shot: This *definitely* needs to be called `WrapError`

src/util.ts
```diff
@@ -421,3 +421,56 @@ interface IUploadResult {
...
+/**
+ * Returns true if `obj` is subtype of at least one of the given types.
+ */
+export function instanceofsome(obj: object, types: Type[]): boolean {
+    return types.some((type) => obj instanceof type);
+}
+
+/**
+ * Append the old error message to the new one and keep its stack trace.
+ * Example:
```
Half-Shot: JsDoc has `@example`

src/util.ts
```diff
@@ -421,3 +421,56 @@ interface IUploadResult {
...
+ */
+export function guildAndChannelOf(remoteRoom: RemoteRoom): [string, string] {
+    const ROOM_NAME_PARTS = 2;
+    return remoteRoom.roomId.substr("_discord".length).split("_", ROOM_NAME_PARTS);
+}
+
+/**
+ * Returns true if the given event is a bot command.
+ */
+export function isBotCommand(event: IMatrixEvent): boolean {
```
Half-Shot: We only really use this in one class, so I'd make it a private function of that class.

src/util.ts
```diff
@@ -421,3 +421,56 @@ interface IUploadResult {
...
+    newError.message += ":\n" + appendMsg;
+    return newError;
+}
+
+/**
+ * Returns the guild and channel of the given remote room extracted from its ID.
+ * @param remoteRoom The room from which to get the guild and channel.
+ * @returns (guild, channel)-tuple.
+ */
+export function guildAndChannelOf(remoteRoom: RemoteRoom): [string, string] {
```
Half-Shot: I can only see this being used once, and tbh it's too small in my opinion to make a function in it's own right.

test/test_matrixeventprocessor.ts
```diff
@@ -17,6 +17,7 @@ limitations under the License.
 import * as Chai from "chai";
 import * as Discord from "discord.js";
 import * as Proxyquire from "proxyquire";
+import { unstable } from "matrix-appservice-bridge";
```
Half-Shot: same as before

src/matrixeventprocessor.ts
```diff
@@ -106,38 +119,32 @@ export class MatrixEventProcessor {
...
             try {
                 await this.HandleEncryptionWarning(event.room_id);
                 return;
             } catch (err) {
-                throw new Error(`Failed to handle encrypted room, ${err}`);
+                throw wrap(err, unstable.EventNotHandledError, `Failed to handle encrypted room, ${err}`);
             }
         } else {
-            log.verbose("Got non m.room.message event");
+            throw new unstable.EventUnknownError("Got non m.room.message event");
```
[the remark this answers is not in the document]
V02460: This is the most exciting bridge error to try out in the wild and one that is quite useful for the user in the end.

I think it will fire for too many events initially and it might become spammy. My current idea is to special case the most common events in a room for which it would fire (especially non-messaging related events), so that it is not annoying, but still catches unknown stuff. (I remember writing that as a more articulated answer to Travis over at the MSC, but I didn't find it at first look…)

I also don't think that it might DoS the bridge, if that is your concern. In contrast to the discussion we had about `m.event_too_old` where events accumulated over many hours might flood in at once, this one will at most spawn as often as the number of incoming events.

src/matrixeventprocessor.ts
```diff
@@ -72,11 +77,19 @@ export class MatrixEventProcessor {
...
+     * @param request Request object containing the event for which this callback is called.
+     * @param context The current context of the bridge.
+     * @throws {unstable.EventNotHandledError} When the event can finally not be handled.
+     */
+    public async OnEvent(request: Request, context: BridgeContext): Promise<void> {
         const event = request.getData() as IMatrixEvent;
         if (event.unsigned.age > AGE_LIMIT) {
-            log.warn(`Skipping event due to age ${event.unsigned.age} > ${AGE_LIMIT}`);
-            return;
+            throw new unstable.EventTooOldError(
```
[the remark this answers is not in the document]
V02460: It doesn't scale, so it's bad. Your decision if we keep it until we implement the solution emerging from the MSC or throw it out and adapt the final solution.

src/util.ts
V02460: It's not so much class related, thus I made it a module-scoped function. (It would be best to have it as method for the (hypothetical) MatrixMsgEvent class, but it's not a class and JS/TS don't have nice ways to do that extension anyways :disappointed: )

src/util.ts
V02460: Made it a function with module scope.

src/matrixeventprocessor.ts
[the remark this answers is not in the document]
Half-Shot: I'd comment it out for now, just because we might potentially do a release before it's MSCed

src/matrixeventprocessor.ts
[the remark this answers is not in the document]
Half-Shot: Hmm, fair point. I agree it's probably not world ending.

src/discordas.ts
```diff
@@ -209,3 +226,73 @@ async function run(port: number, fileConfig: DiscordBridgeConfig) {
...
+    const entries = await store.roomStore.getEntriesByMatrixId(roomId);
+
+    return {
+        rooms: entries[0] || {},
+        senders: {},
+        targets: {},
+    };
+}
+
+class NotReadyError extends Error {
```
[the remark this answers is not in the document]
V02460: Yeeesss....why? :sweat_smile: :stuck_out_tongue_closed_eyes:

Half-Shot:
This looks good to me! Happy to merge?

V02460:
Yes, please go on! :smiley:

Issue #520, opened 2019-07-01
M->D file sending


Sorunome:
If the file is an image and it is too large, try to embed it instead of sending a link

Sorunome:
Done in #521

Issue #517, opened 2019-07-01
puppet disables embeds


Sorunome:
it makes sense for normal accounts, but even if you puppet bot accounts, which doesn't make sense

Sorunome:
Done in #521

PR #521, opened 2019-07-01
Various embed-related sending stuff

Sorunome:
handles #520 and #517

### DIFF
--- a/src/bot.ts
+++ b/src/bot.ts
@@ -419,6 +419,19 @@ export class DiscordBot {
             if (!roomLookup.canSendEmbeds) {
                 // NOTE: Don't send replies to discord if we are a puppet user.
                 msg = await chan.send(embed.description, opts);
+            } else if (!botUser) {
+                const sendEmbed = new Discord.RichEmbed();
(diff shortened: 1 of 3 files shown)


Sorunome:
Soru just realised this won't send properly for bot puppets

Sorunome:
There we go, now it does that, too

src/bot.ts
```diff
@@ -414,18 +416,49 @@ export class DiscordBot {
         }
         try {
             this.channelLock.set(chan.id);
-            if (!botUser) {
-                // NOTE: Don't send replies to discord if we are a puppet.
+            if (!roomLookup.canSendEmbeds) {
+                // NOTE: Don't send replies to discord if we are a puppet user.
```
Half-Shot: I've been bitten by this in #522 and we should fix it.

Sorunome: still, bot puppets can send embeds just fine, whichlook more nicely. So `!roomLookup.canSendEmbeds` would have a different fallback

Sorunome:
Handles #522 now, too

Issue #516, opened 2019-07-01
re-try message sending as webhook


Sorunome:
If you puppet a user and the puppet fails to send a message, e.g. due to insufficiant perms, re-try as webhook

Sorunome:
and/or if it is a puppeted acc check if the user is in the room prior to sending

PR #529, opened 2019-07-15
fallback to webhook if can't send message

Sorunome:
Implements #516


Sorunome:
@Half-Shot ping!

auscompgeek:
Why has package-lock.json changed here? :eyes:

PR #530, opened 2019-07-15
Prefer non-discord canonical aliases

### LINKED ISSUE #523: Bridge doesn't use the canonical alias when referencing rooms
1. Bridge a Discord channel to Matrix
2. Set the canonical alias to something custom
3. In a different Discord channel (or even the same), reference the other channel
4. Note that the alias rendered in Matrix is for the portal room, not the bridged room.

Sorunome:
Fixes #523


src/channelsyncroniser.ts
```diff
@@ -169,15 +169,23 @@ export class ChannelSyncroniser {
...
             rooms = await this.GetRoomIdsFromChannel(channel);
         } catch (err) { } // do nothing, our rooms array will just be empty
+        let fallbackAlias = "";
         for (const room of rooms) {
             try {
                 const al = (await this.bridge.getIntent().getClient()
                     .getStateEvent(room, "m.room.canonical_alias")).alias;
                 if (al) {
-                    return al; // we are done, we found an alias
+                    if (al.startsWith("#_discord_")) {
```
Half-Shot: This feels horribly hacky

Sorunome: We do similar things all over already. Once we switch to the matrix-js-bot-sdk we can use `isNamespacedAlias` or whatever that was called instead

Half-Shot:
Much :)

PR #535, opened 2019-07-17
Fix webhook usersync issues

### LINKED ISSUE #534: webhook avatar
investigate if webhook avatars don't bridge correctly, didn't for one instance here

Sorunome:
fixes #534

### DIFF
--- a/src/usersyncroniser.ts
+++ b/src/usersyncroniser.ts
@@ -233,7 +233,7 @@ export class UserSyncroniser {
         let mxidExtra = "";
         if (webhookID) {
             // no need to escape as this mxid is only used to create an intent
-            mxidExtra = `_${new MatrixUser(`@${webhookID}`).localpart}`;
+            mxidExtra = `_${new MatrixUser(`@${discordUser.username}`).localpart}`;
         }
         const userState: IUserState = Object.assign({}, DEFAULT_USER_STATE, {
             id: discordUser.id,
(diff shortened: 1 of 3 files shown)


src/usersyncroniser.ts
```diff
@@ -233,10 +233,10 @@ export class UserSyncroniser {
         let mxidExtra = "";
         if (webhookID) {
             // no need to escape as this mxid is only used to create an intent
-            mxidExtra = `_${new MatrixUser(`@${webhookID}`).localpart}`;
+            mxidExtra = `_${new MatrixUser(`@${discordUser.username}`).localpart}`;
```
Half-Shot: So what was wrong with webhookID, it's not clear to me?

src/usersyncroniser.ts
```diff
@@ -233,10 +233,10 @@ export class UserSyncroniser {
         let mxidExtra = "";
         if (webhookID) {
```
Half-Shot: if we are no longer using the ID, can this be a boolean instead?

src/usersyncroniser.ts
Sorunome: the problem was that the same webhook can have different names and we don't want username changes constantly. So we fixed that by appending the name to the ID. Only, somehow here, the webhook id instead of the name was appended, which caused problems

Half-Shot: Hm, okay. Can this be commented above?

Sorunome: done

src/usersyncroniser.ts
Sorunome: done

Half-Shot:
I guess this is fine :smile:

PR #544, opened 2019-08-13
Move parsers to matrix-discord-parser

### LINKED ISSUE #459: Build MatrixMessageProcessor for each time an M->D message is processed
because currently there is a `this.prop` thingy  that is been passed to it which contains data like room ID etc. That would mean that that part isn't threadsafe

Sorunome:
This moves the parsers to [matrix-discord-parser](https://github.com/matrix-discord/matrix-discord-parser) for less maintenance of the codebase in the future.

This also fixes #459 due to the threadsafe nature of matrix-discord-parser


Sorunome:
Should some of the tests for matrix message parsing be removed as the other repo covers them already?

Half-Shot:
>Should some of the tests for matrix message parsing be removed as the other repo covers them already?

Yes, if you are 100% sure that the test is covered elsewhere. I'm very happy to reduce the number of tests we need.

PR #548, opened 2019-08-18
Drop requirement for multiple database connectors to be installed

### LINKED ISSUE #400: tsc won't work without better-sqlite

Half-Shot:
Depends on #454 
Fixes #400 

This removes the requirement for `better-sqlite3` and `pg-promise` to be installed together when running the bridge. They will still be required for compiling the project. In addition, this corrects the problem where `source-map-support` was mistakenly included in devDependencies even though it is used at runtime.


Half-Shot:
I think this didn't work from memory, so we will need a different approach.

PR #568, opened 2019-11-01
Fix a minor typo in docs/puppeting.md

ssweeny:
Move a comma to the right spot.


Half-Shot:
Thanks, looks like we broke our linting >:(. but this is fine.

PR #570, opened 2019-11-10
Clarify config.yaml must be in root app folder

MurzNN:
Via current description often users copy `config/config.sample.yaml` file to same folder, so it will placed in `config/config.yaml` instead of `config.yaml`. Will be good to clarify, that it must be placed in root app folder.


Half-Shot:
Sure, works for me. Thanks!

Half-Shot:
buttons

Half-Shot:
ugh

PR #576, opened 2019-12-10
Allow blacklisting mxids

mrjohnson22:
Add new config setting 'bridge.userBlacklist', a list of mxids, to
prevent any messages sent by a blacklisted mxid from being bridged.


mrjohnson22:
No new tests fail when running `npm run test`. I can write a new test for this change if you want, but I won't be able to get to it for at least few days.

src/matrixeventprocessor.ts
```diff
@@ -137,6 +137,10 @@ export class MatrixEventProcessor {
             return;
         } else if (event.type === "m.room.message" || event.type === "m.sticker") {
             log.verbose(`Got ${event.type} event`);
+            if (this.config.bridge.userBlacklist.includes(event.sender)) {
```
Sorunome: shouldn't this blacklist be applied further up, to also prevent messages on `m.room.member` and thelike? When doing that, what if that user redacts someone elses message, should that redaction be bridged?

Half-Shot: I think we would want to drop any events _sent by_ this user, if we're implementing a true blacklist.

Sorunome: Then we would also need to drop the avatar + username changes in `m.room.member`, though

mrjohnson22: I was worried that ignoring things like invite events would corrupt the state of the bridge's view of a room somehow. Thinking about it some more, though, if all state events pertaining to a blacklisted user are ignored, there should be no problem, as that user will never have a chance to affect the state of the room.

Ignoring redactions is a good idea too.

How about ignoring the following:

* All state events related to a blacklisted user (state_key = user)
* All events sent by a blacklisted user (sender = user)

Sorunome: soru means ignoring the "mxid has joined the room on matrix" messages, not necessarily ingorning the invite internally

Sorunome: this method [here](https://github.com/Half-Shot/matrix-appservice-discord/blob/master/src/matrixeventprocessor.ts#L187)

mrjohnson22:
I should point out that the lint & test failures this hits are also encountered in the develop branch. I got rid of the lint problem but not the test failures.

I'll write tests to cover this when I get the chance.

Sorunome:
sorry that it took so long....how about making the blacklist an array of regexes? That way you could easily block everyone from a specific homeserver

mrjohnson22:
That's a good idea. I'' try to get to it soon.

src/config.ts
```diff
@@ -95,6 +95,11 @@ class DiscordBridgeConfigBridge {
     public disableJoinLeaveNotifications: boolean = false;
     public disableInviteNotifications: boolean = false;
     public determineCodeLanguage: boolean = false;
+    public userBlacklist: string[] = [];
+
+    public IsUserBlacklisted(userId: string): boolean {
+        return !!userId && this.userBlacklist.some((regex) => RegExp(`^${regex}$`).test(userId));
```
Sorunome: not sure if adding `^` and `$` here already is a good idea

Sorunome: Like, if you just want to blacklist `:example.org$` or something, it could confuse people that they have to add `.*`

jaller94 (changes requested):
Wording: Let's not use "blacklist"
https://developers.google.com/style/word-list#blacklist

The Slack bridge uses the wording allow and deny.
https://github.com/matrix-org/matrix-appservice-slack/blob/11a9ae875899181f1b9b8b3a23a952805431d5b3/src/AllowDenyList.ts#L3

PR #580, opened 2020-01-04
Add disableInviteNotifications option to drop invitations notifs

fflorent:
This option is complementary to disableJoinLeaveNotifications. A matrix bot may send notifications about users being invite in the room (e.g. IRC users). But we don't want Discord users to be notified (because that creates noises in the discussions).

Also I wonder if the `disableJoinLeaveNotifications` and `disableInviteNotifications` are not redundant in term of use case (I mean I am not sure if users would use different values for both options).


test/test_config.ts
```diff
@@ -61,9 +63,11 @@ describe("DiscordBridgeConfig.applyConfig", () => {
         config.applyEnvironmentOverrides({
             APPSERVICE_DISCORD_BRIDGE_DISABLE_DELETION_FORWARDING: false,
             APPSERVICE_DISCORD_BRIDGE_DISABLE_JOIN_LEAVE_NOTIFICATIONS: true,
+            APPSERVICE_DISCORD_BRIDGE_DISABLE_INVITE_NOTIFICATIONS: true,
```
fflorent: Not sure how that works, but the test seems to pass on my computer.

Sorunome:
Lint failed, would you please fix that? (`npm run lint`). Other than that it looks good!

fflorent:
@Sorunome fixed, thanks!

Half-Shot:
I wonder why the formatting tests are failing :thinking:

fflorent:
@Half-Shot Oh! Is this something I am supposed to fix?

auscompgeek:
It looks like this PR is based on master. Needs to be based on develop, methinks.

fflorent:
@auscompgeek fixed, good catch, thanks!

Sorunome:
This looks good, thank you!

PR #581, opened 2020-01-25
Update packages to make audit happier

Half-Shot:


Sorunome:
Made opsolete in #582

PR #585, opened 2020-02-03
Added discord-markdown to the dependencies

dhghf:


Sorunome:
This is 1st the wrong version and 2nd it is included in matrix-discord-parser with the correct version

Issue #590, opened 2020-03-16
Room avatars are uploaded without MIME type


auscompgeek:
https://github.com/Half-Shot/matrix-appservice-discord/blob/7f84939226fd65ac2e5e265f120c5eda6d9959a0/src/channelsyncroniser.ts#L290-L297

PR #592, opened 2020-03-17
Fix room and user avatars

auscompgeek:
Fixes: #590 (room avatars not rendering in Riot)
Also fixes user avatars not rendering in Riot


Sorunome:
Awesome, thank you!

PR #596, opened 2020-04-10
Upload attachments and user avatars as ghosts

auscompgeek:


auscompgeek:
I'm not going to have time to fix the tests any time soon. It'd be great if anyone else had an opportunity to pick this up.

PR #605, opened 2020-07-04
Port to ESLint

### LINKED ISSUE #526: Move to use eslint sometime in the future
See https://medium.com/palantir/tslint-in-2019-1a144c2317a9

Half-Shot:
TSLint died off a long time ago, we should be moving to eslint. I've started trying to port our styles across, but it's non-trivial to match up the eslint rules with our tslint rules in some places.

Fixes #526


package.json
```diff
@@ -5,7 +5,7 @@
   "main": "discordas.js",
   "scripts": {
     "test": "npm run-script build && mocha build/test/config.js build/test",
-    "lint": "tslint --project ./tsconfig.json -t stylish",
+    "lint": "eslint -c .eslintrc --max-warnings 0 src/**/*.ts test/**/*.ts",
```
jaller94, line 8: Why was this set to allow 0 warnings? This fails the CI in my branch #637.
Removing all the warnings is unrealistic with the current ESLint config and code.
Furthermore, warnings would behave just like errors.

PR #617, opened 2020-09-27
WIP: start migrating

Sorunome:


Half-Shot:
Thanks soru for attempting to push us into the new world. Naturally I failed to take any notice and ended up doing #626. Hopefully it will all be worth it.

PR #626, opened 2020-10-28
Update bridge to be compatible with the new Discord APIs

Half-Shot:
The bridge is currently **non-functional** due to some API changes on Discord, and we missed the deadline. The bridge is currently on a `v11` API and we need to be on `v12`. We're going to be using @Sorunome 's fork of `better-discord.js` to ensure we can do puppeting.

~~Tests are currently expected to fail.~~ They should now pass, but we're stuck on https://github.com/matrix-discord/matrix-discord-parser/issues/9

This PR also vapes `package-lock.json` in favour of `yarn.lock`, because I was getting quite tired of npm messing up.


Half-Shot:
This is blocked by https://github.com/matrix-discord/matrix-discord-parser/issues/9.

.travis.yml
```diff
@@ -1,6 +1,6 @@
 dist: xenial
 language: node_js
-install: npm ci
+install: npm i
```
auscompgeek, line 3: Shouldn't this be `yarn install` instead, if you're moving to yarn?

src/bot.ts
```diff
@@ -82,6 +83,7 @@ export class DiscordBot {
     /* Handles messages queued up to be sent to matrix from discord. */
     private discordMessageQueue: { [channelId: string]: Promise<void> };
     private channelLock: Lock<string>;
+    private typingTimers: Record<string, NodeJS.Timeout> = {}; // DiscordUser+channel -> Timeout
```
Sorunome, line 86: why not just set synapse timeout to 30s?

src/clientfactory.ts
```diff
@@ -42,11 +42,18 @@ export class DiscordClientFactory {
...
+        });
+
+        const waitPromise = new Promise((resolve, reject) => {
+            this.botClient.once("shardReady", resolve);
+            this.botClient.once("shardError", reject);
         });
 
         try {
-            await this.botClient.login(this.config.botToken);
+            await this.botClient.login(this.config.botToken, true);
```
Sorunome, line 53: You might want to store if tokens are user or bot tokens in the db. After a ready you can fetch your own status and store that in the db for existing ones set without explicitly specifying if it is bot or user

Sorunome (commented):
Seems mostly sensible

src/clientfactory.ts
```diff
@@ -42,11 +42,18 @@ export class DiscordClientFactory {
         this.botClient = new DiscordClient({
             fetchAllMembers: true,
```
Sorunome, line 43: Having this set to `true` was causing issues with t2bot not starting up.

src/bot.ts
Half-Shot: TIL that's a thing.

src/clientfactory.ts
Sorunome: The problem basically is that members is a priviliged intent and as soon as your bot is in > 100 guilds this setting will screw up the bot, if the bot owner does not go through all the hoops to get it whitelisted

src/clientfactory.ts
```diff
@@ -42,11 +42,18 @@ export class DiscordClientFactory {
...
             messageCacheLifetime: 5,
-            sync: true,
```
Sorunome, line 45: You didn't specify which itnents to use at all, so this won't work after nov 7th

Sorunome: `new Discord.Client({ ws: { intents: Discord.Intents.NON_PRIVILEGED }})`

You might want to add a setting for if it'll use priveleged or unpriveleged intents

src/clientfactory.ts
Half-Shot: We only ever store user tokens in the DB. The only bot token is in the config. If we ever changed this we would add appropriate DB stuffs.

src/clientfactory.ts
```diff
@@ -95,6 +101,9 @@ export class DiscordClientFactory {
         const client = new DiscordClient({
             fetchAllMembers: true,
```
Sorunome, line 102: `fetchAllMembers: this.config.usePriviledgedIntents`

src/clientfactory.ts
Sorunome: Actually, a user might want to puppet a bot token.

config/config.sample.yaml
```diff
@@ -38,6 +38,9 @@ auth:
   # This MUST be a string (wrapped in quotes)
   clientID: "12345"
   botToken: "foobar"
+  # You must enable "Privileged Gateway Intents" in your bot settings on discord.com (e.g.g https://discord.com/developers/applications/12345/bot)
+  # for this to work
+  usePriviledgedIntents: false
```
auscompgeek, line 43: typo?

package.json
```diff
@@ -36,15 +36,16 @@
   },
   "homepage": "https://github.com/Half-Shot/matrix-appservice-discord#readme",
   "dependencies": {
+    "better-discord.js": "https://github.com/Sorunome/better-discord.js.git#b5a28499899fe2d9e6aa1aa3b3c5d693ae672117",
```
pacien, line 39:
```suggestion
    "better-discord.js": "git://github.com/Sorunome/better-discord.js.git#b5a28499899fe2d9e6aa1aa3b3c5d693ae672117",
```

While Yarn seems to gladly accept `https` for git repository, other tools like yarn2nix which is used to generate the Nix package require the right protocol to be set to properly download the dependency.

([It otherwise tries to download the page, confusing it with a tarball](https://github.com/NixOS/nixpkgs/blob/b22ab697faf5b8140b5eb1952220532d88492669/pkgs/servers/matrix-appservice-discord/yarn-dependencies.nix#L676-L683))

.travis.yml
```diff
@@ -1,25 +1,22 @@
...
   include:
-      # Lint doesn't need to build
     - stage: lint
-      script: npm run lint
+      script: yarn lint
       node_js: "12"
     - stage: unit tests
-      # Test already builds
-      script: npm run test
+      script: yart test
```
pacien, line 15: Typo?

```suggestion
      script: yarn test
```

.travis.yml
```diff
@@ -1,25 +1,22 @@
...
       node_js: "10"
     - node_js: "12"
```
pacien, line 17: The test job fails with NodeJS 10 because better-discord 12 requires NodeJS ≥12.

```
error better-discord.js@12.3.1: The engine "node" is incompatible with this module. Expected version ">=12.0.0". Got "10.23.0"
```

The new release might be the occasion to drop compatibility with this old version of Node.

```suggestion
      node_js: "12"
```

__The readme file needs to be updated accordingly as well.__

src/clientfactory.ts
Half-Shot: Yes, I think that falls under "If we ever changed this we would add appropriate DB stuffs."

Half-Shot:
Things seem to be sufficiently stable. going to merge.

PR #639, opened 2020-12-02
Add /health HTTP endpoint

jaller94:
For a remote process (e.g. Ansible) to check whether or not the bridge is generally listening to a port, I'm proposing a `/health` endpoint. If the bridge does not respond, a monitoring process may alert its administrators and restart the bridge.

This is inspired by the `/health` and `/ready` endpoints that can be found in [this Slack bridge](https://github.com/matrix-org/matrix-appservice-slack/blob/34833214c3c300de2e6cd5bf16318b27a7a2cffa/src/Main.ts#L1471-L1477).


src/discordas.ts
```diff
@@ -176,6 +176,11 @@ async function run() {
     const roomhandler = discordbot.RoomHandler;
     const eventProcessor = discordbot.MxEventProcessor;
 
+    // tslint:disable-next-line:no-any
+    appservice.expressAppInstance.get("/health", (_, res: any) => {
+        res.status(201).send("");
```
auscompgeek: Is there a particular reason to return [201](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/201)?

jaller94: Only that the Slack bridge does it this way... `Created` does not make any sense at all.
Lets change it to 200.

auscompgeek: If there was no reason to use 201, surely 204 No Content is better?

jaller94: I think 204 is for omitting even though there may be content.
It expects the server to send an `ETag` header to indicate that the content didn't change.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/204

Responding 200 without content actually means "There is no content".

src/discordas.ts
```diff
@@ -176,6 +176,11 @@ async function run() {
     const roomhandler = discordbot.RoomHandler;
     const eventProcessor = discordbot.MxEventProcessor;
 
+    // tslint:disable-next-line:no-any
```
Half-Shot: We can't import express types?

Half-Shot (approved):
Fine, but the any seems odd?

src/discordas.ts
auscompgeek: I think that's saying you don't need to send other cache-related headers for a 204 to be cacheable (indeed, reading the RFC that's exactly what it's saying).

src/discordas.ts
jaller94: Added in a65bc62.

Issue #641, opened 2020-12-07
After editing on Matrix, the posted edit message on Discord has an embed


90-008:
![shot_2020_12_07_19_51](https://user-images.githubusercontent.com/19897088/101379680-984fa000-38c5-11eb-9955-31a8a3972ff0.png)

Similar thing also happens with mentions:
![shot_2020_12_07_19_52](https://user-images.githubusercontent.com/19897088/101379819-c6cd7b00-38c5-11eb-9f47-9f9e336e9d58.png)

Is it possible to remove these embeds so that they don't take up space?

Sanqui:
Apparently the bridge has a `suppress_embeds` flag, but it's not set on the running instance.  Please set it, these embeds are really annoying and disruptive.

timokoesters:
I just messed around with this a lot and it looks like links like this work and do not trigger a preview:
```
[THIS WORKS](https://a:a@discord.com/channels/770973140155105300/770973140155105304/791022097680826379)
```

shmerl:
@jonaharagon: is there any blocker to merge the fix?

skybird23333:
`<link>` removes link preview embeds on discord.

AnonyPla:
Please merge the pull request

PR #643, opened 2020-12-08
Update user-facing discordapp refs to discord

auscompgeek:
(the CDN is still cdn.discordapp.com)


auscompgeek:
This may fix, possibly, #641.

PR #646, opened 2020-12-09
Remove unused parameter `provisioner` from MatrixRoomHandler

jaller94:
TypeScript told me that this was unused code and I couldn't find it being used anywhere.

:warning: If this bridge/file is used as a library, this is API-breaking. That would require a new major version.


Sorunome:
This file isn't / shouldn't be used as a library, so it is ok to merge

PR #661, opened 2021-01-02
Fix presence status error, fix client init

karolyi:
I've been pinging you on the matrix channel.

The intent fixes are for when the `usePrivilegedIntents` is set to `true` in the config, and the client passed wrong flags so no channels, no messages, no nothing would come through the API. I, as a newcomer to this module and the discord API, worked on it a couple days until I managed to figured this out.

The status fix is for the presence updates, as `null` could be passed which the matrix api doesn't tolerate and gives an error in return. An empty string when `null` would be passed is accepted.

### DIFF
--- a/src/clientfactory.ts
+++ b/src/clientfactory.ts
@@ -69,7 +69,7 @@ export class DiscordClientFactory {
             fetchAllMembers: false,
             messageCacheLifetime: 5,
             ws: {
-                intents: this.config.usePrivilegedIntents ? Intents.ALL : Intents.NON_PRIVILEGED,
+                intents: Intents.NON_PRIVILEGED,
             },
         });
 
@@ -102,7 +102,7 @@ export class DiscordClientFactory {
             fetchAllMembers: false,
             messageCacheLifetime: 5,
             ws: {
-                intents: this.config.usePrivilegedIntents ? Intents.ALL : Intents.NON_PRIVILEGED,
+                intents: Intents.NON_PRIVILEGED,
             },
         });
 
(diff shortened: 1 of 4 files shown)


src/clientfactory.ts
```diff
@@ -69,7 +69,7 @@ export class DiscordClientFactory {
             fetchAllMembers: false,
             messageCacheLifetime: 5,
             ws: {
-                intents: Intents.NON_PRIVILEGED,
+                intents: this.config.usePrivilegedIntents ? Intents.ALL : Intents.NON_PRIVILEGED,
```
Sorunome: This is typically for usertokens and was purposefully omitted from `this.config.usePrivilegedIntents`

src/clientfactory.ts
```diff
@@ -102,7 +102,7 @@ export class DiscordClientFactory {
...
```
Sorunome: same here

Sorunome (changes requested):

src/clientfactory.ts
karolyi: The module should have `this.config` at this point, so I guess it shouldn't matter. It's for consistency.

src/clientfactory.ts
karolyi: https://github.com/Half-Shot/matrix-appservice-discord/pull/661#discussion_r550872147

src/clientfactory.ts
Sorunome: That's not the point. The client objects created here are *not* the main bot one. In fact, they are only ever used to send messages, and not read them. They are for if someone has configured a puppet. There is no usecase for priviliged intents for these at all, and them being different tokens does not mean you have priviliged intents enabled for the ones here.

karolyi: should be okay now

Sorunome (approved):
Looks good, thanks a lot!

PR #664, opened 2021-01-10
Update matrix-bot-sdk to 0.5.11 in package.json

vranki:
This is tiny but extremely important change.

With matrix-bot-sdk 0.5.4 the maximum size for HTTP request from Synapse to bridge is 100kb. When the first request exceeding this is sent, the bridge will fail and never recover. From user perspective bridge stops receiving anything from Matrix side. Messages from matrix to discord stop being relayed and bot stops responding to commands or invites.

I would suggest creating a new release due to importance of this change.

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
     "escape-string-regexp": "^4.0.0",
     "js-yaml": "^3.14.0",
     "marked": "^1.2.2",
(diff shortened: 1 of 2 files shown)


auscompgeek:
The lockfile needs an update.

PR #666, opened 2021-01-15
Fix typo with options.reg

cmdruid:
The '-r' flag value is stored under the alias 'registration', but is later passed along to ToolsHelper.getToolDependencies() as 'options.reg' which is undefined. This causes the internal default of './discord-registration.yaml' to always be used, instead of what the user defined. Should be fixed by changing options.reg to options.registration.

### DIFF
--- a/tools/adminme.ts
+++ b/tools/adminme.ts
@@ -95,7 +95,7 @@ if (!options.userid) {
     process.exit(1);
 }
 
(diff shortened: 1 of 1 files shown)


auscompgeek (approved):

Sorunome:
Is fixed via #667 now?

PR #671, opened 2021-01-22
Fixed some issues I had with metrics

tommy-mor:
relevant line:  [here](https://github.com/Half-Shot/matrix-appservice-discord/blob/2f37665ae8c5ee1b8345a0165e100bf9972e2c64/config/config.sample.yaml#L109)

Not sure what the issue is with the change in metrics.ts, i'm new to typescript, but this fixed it for me.

### DIFF
--- a/src/discordas.ts
+++ b/src/discordas.ts
@@ -163,7 +163,7 @@ async function run(): Promise<void> {
         storage: store,
     });
 
-    if (config.metrics.enable) {
+    if (config.metrics.enabled) {
         log.info("Enabled metrics");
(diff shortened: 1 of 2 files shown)


jaller94 (changes requested):
The flag has been called `metrics.enable` for over two years. It was falsely added as `enabled` to TypeScript, config and schema only a few months ago.

For being backwards-compatible with old, correct config files I advise against this PR.
It may be good to support both `enable` and `enabled` as `logging.files[]` also has a property `enabled`.

tommy-mor:
ok cool! thanks for fixing it (properly) with your pull. Any idea about the second change to the imports?

PR #677, opened 2021-02-02
Lint all .ts files in the folders src, test and tools

### LINKED ISSUE #676: Linter: The files src/*.ts and test/*.ts are not linted

jaller94:
Fixes #676 

The files have many errors and warnings with the current rules, so the rules may need to be adapted.


package.json
```diff
@@ -5,7 +5,7 @@
   "main": "discordas.js",
   "scripts": {
     "test": "mocha -r ts-node/register test/config.ts test/test_*.ts test/**/test_*.ts",
```
xenofem:
```suggestion
    "test": "mocha -r ts-node/register test/config.ts 'test/**/test_*.ts'",
```
With properly quoted globs, we can simplify this command too.

.eslintrc
```diff
@@ -27,21 +27,38 @@
...
+            },
+        ],
+        "prefer-template": "off",
         // Disable these as they were introduced by @typescript-eslint/recommended
         "@typescript-eslint/no-use-before-define": "off",
         "@typescript-eslint/no-inferrable-types": "off",
         "@typescript-eslint/member-delimiter-style": "off",
         "@typescript-eslint/no-unused-expressions": "off",
-        "@typescript-eslint/interface-name-prefix": "off"
+        "@typescript-eslint/interface-name-prefix": "off",
```
MightyCreak: The JSON syntax doesn't allow trailing `,` at the end of a block

.eslintrc
```diff
@@ -27,21 +27,38 @@
...
         "no-bitwise": "error",
         "no-console": "error",
         "no-debugger": "error",
-        "prefer-template": "error",
+        "@typescript-eslint/naming-convention": ["error",
+            {
+                "selector": "default",
+                "format": ["camelCase", "PascalCase", "snake_case", "UPPER_CASE"],
+                "leadingUnderscore": "allow",
+                "trailingUnderscore": "forbid",
```
MightyCreak: The JSON syntax doesn't allow trailing , at the end of a block

.eslintrc
```diff
@@ -50,7 +67,21 @@
...
             "rules": {
                 "@typescript-eslint/no-empty-function": "off",
-                "@typescript-eslint/no-explicit-any": "off"
+                "@typescript-eslint/no-explicit-any": "off",
+                "@typescript-eslint/no-unused-vars": ["warn",
+                    {
+                        "varsIgnorePattern": "^_"
+                    }
+                ],
+                "@typescript-eslint/no-var-requires": "off",
```
MightyCreak: The JSON syntax doesn't allow trailing , at the end of a block

MightyCreak (changes requested):

MightyCreak (commented):
@Half-Shot Overall, this PR looks good to me! I always think stricter linting is better in order to simplify the contribution.

The only parts I'm unsure, as it depends on the code convention you would like to enforce, are: the function arguments on multiple lines (though right now there are different conventions) and the opening square bracket for arrays.

What do you think?

.eslintrc
jaller94: I think `.jsonrc` is parsed as https://json5.org/, which does allow trailing commas.

MightyCreak: Indeed, since it's not pure JSON, they can use a better parser... As long as it works, I'm fine with it ;)

PR #678, opened 2021-02-02
Metrics.enable: Correct config sample and schema

jaller94:
In #640 I mistakenly added the config flag as `enabled` instead of `enabled`.

Not that another flag for logging is already called `enabled`. It may be worth supporting both tenses.


jaller94:
Similar to #671 which wants to rename the flag.

PR #679, opened 2021-02-04
Do not use default

Half-Shot:
Apparently this is not the way. I do not know why these types worked at all.

### DIFF
--- a/src/metrics.ts
+++ b/src/metrics.ts
@@ -15,7 +15,7 @@ limitations under the License.
 */
 /* eslint-disable max-classes-per-file, @typescript-eslint/no-empty-function */
 
-import { Gauge, Counter, Histogram, default as promClient } from "prom-client";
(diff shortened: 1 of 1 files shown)


src/metrics.ts
```diff
@@ -69,57 +69,58 @@ export class PrometheusBridgeMetrics implements IBridgeMetrics {
...
 
     public init(as: Appservice, config: DiscordBridgeConfigMetrics) {
-        promClient.collectDefaultMetrics();
+        collectDefaultMetrics();
         // TODO: Bind this for every user.
         this.httpServer = http.createServer((req, res) => {
             if (req.method !== "GET" || req.url !== "/metrics") {
                 res.writeHead(404, "Not found");
                 res.end();
+                return;
```
jaller94, line 78: Nice catch!

jaller94:
Looks good to me.
I guess they may have had a breaking change of how they export modules within the 12.x releases..
`default` is a special thingy.

PR #680, opened 2021-02-04
fix: use default permissions when not set explicitly

timokoesters:
https://matrix.org/docs/spec/client_server/r0.6.1#m-room-power-levels

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -350,7 +350,7 @@ export class Util {
         if (res && res.users_default) {
             haveLevel = res.users_default;
         }
-        if (res && res.users && res.users[userId] !== undefined) {
+        if (res && res.users && typeof res.users[userId] === "number") {
             haveLevel = res.users[userId];
         }
         return haveLevel >= requiredLevel;


src/util.ts
```diff
@@ -347,8 +347,8 @@ export class Util {
         }
 
         let haveLevel = 0;
-        if (res && res.users_default) {
-            haveLevel = res.users_default;
+        if (res && res.users_default || 0) {
```
auscompgeek: This change does nothing.

timokoesters: My bad, I will remove the || 0 in this line again

src/util.ts
```diff
@@ -347,8 +347,8 @@ export class Util {
...
+            haveLevel = res.users_default || 0;
         }
         if (res && res.users && res.users[userId] !== undefined) {
```
Sorunome:
```suggestion
        if (res && res.users && typeof res.users[userId] === "number") {
```

timokoesters: To be honest I don't know why it was failing before and it doesn't seem to fail anymore even without my commit, so I just forced pushed just this change

PR #690, opened 2021-03-30
Fix incorrect adminme example

xenofem:
The room id is given with the `-m` flag, not `-r`.


auscompgeek (approved):

xenofem:
Oh, looks like this is a duplicate of #672, sorry for not noticing!

PR #691, opened 2021-04-02
Quote globs in npm scripts so globstars are expanded properly

xenofem:
https://medium.com/@jakubsynowiec/you-should-always-quote-your-globs-in-npm-scripts-621887a2a784

Without proper globstar handling, linting was only getting applied to files in subdirectories of `src` and `test`, not files directly inside `src` and `test`.


xenofem:
And this is a duplicate of #677 , I'll try to be better about checking these ^^;;

PR #703, opened 2021-05-24
Add reply to feature-matrix

JulianGro:
Both Martix and Discord have a reply function which basically quotes a message when replying to it.

When replying to a message from Discord inside of Matrix, the reply shows up inside Discord.

When replying to a message from Matrix inside of Discord, the reply does not show up inside Matrix.

Example:
1Auswahl_895
1Auswahl_896


MightyCreak:
This is just a README edit and the changes are valid. This PR can be merge IMHO.

Also, it will be fixed with #704

MightyCreak (approved):

Issue #631, opened 2020-11-20
Discord reply support


StepS-:
The bridge will need to account for Discord users replying to other messages. (no source message/sender will currently be shown)

auscompgeek:
discord.js will first need to gain inline reply support of course - it doesn't look like they even have a tracking issue for it yet.

auscompgeek:
Looks like there's a PR though: discordjs/discord.js#4874

auscompgeek:
So it looks like it's possible currently (without updating the library) to determine whether a Discord message is a reply by looking at `Message#reference`. So we can translate Discord replies to Matrix replies.

It would be good to send Discord replies at some point though.

auscompgeek:
The discord.js PR finally got merged :tada:

Would it be worth merging the upstream changes into better-discord.js and making a prerelease for the bridge?

HansJK:
Any updates whether the bridge is going to support Discord replies?

rollingmoai:
- [ ] Discord->Matrix: #706 (Fixed in #704)
- [ ] Matrix->Discord: #693

While Matrix->Discord replies are a bit trickier, as webhooks do not support replies, one can use buttons instead. See [Connections Bot](https://combot.xyz/)'s implementation:
![image](https://user-images.githubusercontent.com/75471378/188256072-55a04321-7501-4358-bd4e-fc5416150a00.png)

PR #704, opened 2021-05-25
Discord To Matrix inline replies

cernodile:
This should largely reduce confusion when people are talking over the bridge as the context often gets lost when people reference posts unrelated to current conversations.

End result: 
image

Edit: the previously mentioned PR is no longer necessary.

### DIFF
--- a/src/bot.ts
+++ b/src/bot.ts
@@ -1005,6 +1005,20 @@ export class DiscordBot {
                     formatted_body: result.formattedBody,
                     msgtype: result.msgtype,
                 };
+                if (msg.reference) {
+                    const storeEvent = await this.store.Get(DbEvent, {discord_id: msg.reference?.messageID})
+                    if (storeEvent && storeEvent.Result)
+                    {
+                        while(storeEvent.Next())
+                        {
+                            sendContent["m.relates_to"] = {
+                                "m.in_reply_to": {
+                                    event_id: storeEvent.MatrixId.split(";")[0]
+                                }
+                            };
+                        }
+                    }
+                }
                 if (editEventId) {
                     sendContent.body = `* ${result.body}`;
                     sendContent.formatted_body = `* ${result.formattedBody}`;


steewbsd:
Can confirm it works in my tests.

BBaoVanC:
Is there anything blocking this from being merged?

timokoesters:
Replies will usually have the user id and message content they replied to as the raw body. I think this current implementation will not ping a matrix user if a discord user replies to them.

timokoesters:
I realized you also need to put mx-reply into the formatted body or else clients like Element Android will not show the reply properly

cernodile:
I am aware of this. I did originally open another PR in the other repository (look at history of this PR), but was told that it's not necessary. I was advised to use this method only as it is the right way per spec.

You should bring the latter issue up with Element Android folks.

timokoesters:
I spoke to Element Android devs: 

> "EA has some bugs, but according to [the spec](https://matrix.org/docs/spec/client_server/r0.6.1#rich-replies) clients (and so bridges I guess (?)) MUST provide fallback: Clients that do support rich replies MUST provide the fallback format on replies"

cernodile:
> I spoke to Element Android devs:
> 
> > "EA has some bugs, but according to [the spec](https://matrix.org/docs/spec/client_server/r0.6.1#rich-replies) clients (and so bridges I guess (?)) MUST provide fallback: Clients that do support rich replies MUST provide the fallback format on replies"

In this case, there's either a) reviving the PR on other repository (unlikely to be merged) or b) hackily parsing a html fallback here. This PR does seem stagnant at this point, though. I'll look for a solution if the repository shows signs of life again.

src/bot.ts
```diff
@@ -1005,6 +1005,20 @@ export class DiscordBot {
                     formatted_body: result.formattedBody,
                     msgtype: result.msgtype,
                 };
+                if (msg.reference) {
+                    const storeEvent = await this.store.Get(DbEvent, {discord_id: msg.reference?.messageID})
+                    if (storeEvent && storeEvent.Result)
+                    {
+                        while(storeEvent.Next())
+                        {
+                            sendContent["m.relates_to"] = {
```
kevincox, line 1014: `m.relates_to` is also set below with an override like this. The object structures seem incompatible but I think we should consider one of the following:

1. Merge the `m.relates_to` objects.
2. Put a comment explaining that edit relations will override reference relations and why that is correct.

turt2live:
This has been [deployed](https://matrix.to/#/!eTLgPMtcqcAGwtXDRF:t2bot.io/$x_gan3JS8NFI8_AnklbXhSi7gJ15lfSBjb5tGJXfxHU?via=t2l.io&via=matrix.org&via=envs.net) on t2bot.io

jonaharagon:
This seems to have occasional trouble replying to messages that are a bit further back in the timeline. I didn't run into this issue hosting my own bridge with this code, but I'm seeing it on t2bot.io fairly frequently. Maybe this isn't fully functional for much larger installs?

Example (both Spoingus and UltrxLinx are t2bot bridge users):

image

This message's source shows:

```
      "m.in_reply_to": {
        "event_id": "$16408893244317MqvyF:t2bot.io"
      }
```

When the event ID should have been `$xj7T7VRAyTOtg7XqpKKacxlc27UnhuQ3n_N8hmCWQVA`.

turt2live:
in the past that's been because the bridge somehow got confused and bridged the room to two places - visit the t2bot.io help room and we'll get you on the right track there.

puyoxyz:
discord has a reply feature now, maybe update this to use that reply feature instead of an embed ?

cernodile:
> discord has a reply feature now, maybe update this to use that reply feature instead of an embed ?

This is outside the scope of PR and webhooks are unable to use inline reply functionality.

puyoxyz:
> webhooks are unable to use inline reply functionality.

ah, i was unaware of that. nevermind then

theCapypara:
Can this be merged, since it's already deployed to the public  anyway?

rollingmoai:
> webhooks are unable to use inline reply functionality.

@cernodile One can probably use buttons instead. See https://github.com/matrix-org/matrix-appservice-discord/issues/631#issuecomment-1236047430:

> While Matrix->Discord replies are a bit trickier, as webhooks do not support replies, one can use buttons instead. See [Connections Bot](https://combot.xyz/)'s implementation:
image

cernodile:
> > webhooks are unable to use inline reply functionality.
> 
> @cernodile One can probably use buttons instead. See [#631 (comment)](https://github.com/matrix-org/matrix-appservice-discord/issues/631#issuecomment-1236047430):
> 
> > While Matrix->Discord replies are a bit trickier, as webhooks do not support replies, one can use buttons instead. See [Connections Bot](https://combot.xyz/)'s implementation:
> > image

Hello!

Thank you for bringing this to my attention. I think this is currently out of the scope for now as it'd require replacing our current better-discord.js dependency with upstream one (which feasibility would also need to be considered first). I will most likely experiment with this over time on my fork of the project.

Miepee:
@cernodile would it be possible for you to rebase this and adhere to the new contribution guidelines? (signed off, changelog file, tests etc.)

z411:
Since this seems to be stalled I merged this in my instance. It works great, thanks. Except for the ping; what could be done so the user that was replied to gets pinged?

salixor:
Unless I'm mistaken, there's an issue with the following use case :
- A Discord user sends a first message (Event A on Matrix)
- They edit their message (Event B on Matrix, which replaces event A)
- A second Discord user replies to that now edited message (Event C)
- The reply on Matrix says "This event could not be displayed"

This is due to using the wrong event being replied to. Event C tries to reply to Event B, but it should reply to Event A.

Ancurio:
> Unless I'm mistaken, there's an issue with the following use case :
> 
>     * A Discord user sends a first message (Event A on Matrix)
> 
>     * They edit their message (Event B on Matrix, which replaces event A)
> 
>     * A second Discord user replies to that now edited message (Event C)
> 
>     * The reply on Matrix says "This event could not be displayed"
> 
> 
> This is due to using the wrong event being replied to. Event C tries to reply to Event B, but it should reply to Event A.

We also deployed this patch on mapleshrine.eu now; the quoted behavior is a bit unfortunate, but this is still heaps better than discord replies not being communicated at all to matrix (thanks a lot @cernodile!). Is it possible to determine the new event id in case a discord message gets edited?

PR #735, opened 2021-10-01
Unbridge portals

mrjohnson22:
As far as I can tell, these changes allow portal rooms to be unbridged.

I've only tested this in conjunction with #733, but those changes shouldn't be strictly required for these ones to make sense, so I figured I'd split this into its own PR.


jimeh:
This is a feature I'd like, as I have a portal room against a Discord room which is now bridged to another pre-existing Matrix channel, but all messages goes to both, so I'd like to remove the portal room completely.

PR #736, opened 2021-10-01
Config to auto-apply power levels to designated users

mrjohnson22:
This serves as an automated alternative to having to run "adminme" on the same user for every portal room the bridge creates.


Jaykul:
Looks ok. I'm not sure we really want the bridge handling that, but it's optional, so ... fine.

PR #740, opened 2021-10-12
Implement RMAU tracking and optionally blocking the bridge after a RMAU limit is reached

tadzik:
This makes us dependent on matrix-appservice-bridge to make use of
UserActivityTracker and BridgeBlocker, which also bumps matrix-bot-sdk
to a higher required version.

### DIFF
--- a/src/bot.ts
+++ b/src/bot.ts
@@ -1197,7 +1197,7 @@ export class DiscordBot {
         for (const userId of state.changed) {
             await this.store.storeUserActivity(userId, state.dataSet.users[userId]);
         }
-        log.info(`Checking bridge limits (${state.activeUsers} active users)`);
+        log.verbose(`Checking bridge limits (${state.activeUsers} active users)`);
         this.bridgeBlocker?.checkLimits(state.activeUsers);
         MetricPeg.get.setRemoteMonthlyActiveUsers(state.activeUsers);
     }
--- a/src/bot.ts
+++ b/src/bot.ts
@@ -178,8 +178,6 @@ export class DiscordBot {
 
     public GetIntentFromDiscordMember(member: Discord.GuildMember | Discord.PartialUser | Discord.User,
                                       webhookID: string|null = null): Intent {
-        let intent: Intent;
-
         if (webhookID) {
             // webhookID and user IDs are the same, they are unique, so no need to prefix _webhook_
             const name = member instanceof Discord.GuildMember ? member.user.username : member.username;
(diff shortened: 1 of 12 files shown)


src/metrics.ts
```diff
@@ -132,6 +138,18 @@ export class PrometheusBridgeMetrics implements IBridgeMetrics {
         }, REQUEST_EXPIRE_TIME_MS);
         this.httpServer.listen(config.port, config.host);
 
+        this.remoteMonthlyActiveUsers = new Gauge({
+            help: "Current count of remote users active this month",
+            name: "remote_monthly_active_users",
```
Half-Shot: N.B I have a horrible feeling that none of these are namespaced to `bridge_`

src/discordas.ts
```diff
@@ -94,13 +94,15 @@ function setupLogging(): void {
 
     LogService.setLogger({
```
Half-Shot, line 95: Seems okay, did something break?

src/metrics.ts
tadzik: Ah yes, MAB did that for us for free IIRC, and this does not. I'll adjust.

Half-Shot: Oh hum, except I expect people like @turt2live expect the metrics to be named the old crappy way....we might need a config option.

src/bot.ts
```diff
@@ -146,18 +188,40 @@ export class DiscordBot {
...
             }
             // TODO: We need to sanitize name
-            return this.bridge.getIntentForSuffix(`${webhookID}_${Util.EscapeStringForUserId(name)}`);
+            intent = this.bridge.getIntentForSuffix(`${webhookID}_${Util.EscapeStringForUserId(name)}`);
+        } else {
+            intent = this.bridge.getIntentForSuffix(member.id);
         }
-        return this.bridge.getIntentForSuffix(member.id);
+
+        this.userActivity.updateUserActivity(intent.userId);
```
Half-Shot: Do we always use this intent to send a message though? I think we also use it for things like presence setting :s

Half-Shot (commented):
Some thoughts.

src/bot.ts
tadzik: I thought that counts as "remote user activity". I can change it though.

src/metrics.ts
turt2live: (I don't)

src/bot.ts
Half-Shot: I think we decided on just message sending.

src/discordas.ts
tadzik: Yeah, matrix-bot-sdk's API, according to typescript :)

src/bot.ts
tadzik: Changed in 677affc0c9b9c0d016b94cfa62bf7839b5f67e7d. All visible room actions are covered, which is consistent with MAB's behaviour (which grabs all `sendEvent`s iirc).

Half-Shot (approved):
Looks good to me!

PR #746, opened 2021-11-11
Handle invalid Discord tokens gracefully

tadzik:
732be16490814a28ec8e46d441e9498912413bd8 ensures we don't crash on startup, and thus not crashloop if the token has been regenerated or configured incorrectly.

09a36313d186379c32a26079b5c2cfea6f1ccaf9 proposes a way to notify a bridge admin (optional, configurable) about the startup problems.

### DIFF
--- a/src/discordas.ts
+++ b/src/discordas.ts
@@ -13,7 +13,7 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 */
-import { Appservice, IAppserviceRegistration, LogService } from "matrix-bot-sdk";
+import { Appservice, IAppserviceRegistration, LogService, MatrixClient } from "matrix-bot-sdk";
 import * as yaml from "js-yaml";
 import * as fs from "fs";
 import { DiscordBridgeConfig } from "./config";
@@ -29,6 +29,8 @@ import { MetricPeg, PrometheusBridgeMetrics } from "./metrics";
 import { Response } from "express";
 
 const log = new Log("DiscordAS");
+
+const ADMIN_ROOMS_KEY = 'uk.half-shot.discord.admin-rooms';
 
 const commandOptions = [
     { name: "config", alias: "c", type: String },
@@ -215,7 +217,7 @@ async function run(): Promise<void> {
     roomhandler.bindThirdparty();
 
     try {
-        await startDiscordBot(discordbot);
+        await startDiscordBot(discordbot, appservice.botClient, config.bridge.adminMxid);
         log.info("Discordbot started successfully");
     } catch (err) {
         log.error(err);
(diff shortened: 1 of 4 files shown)


src/discordas.ts
```diff
@@ -214,20 +216,67 @@ async function run(): Promise<void> {
...
     try {
         await discordbot.init();
         await discordbot.run();
-        log.info("Discordbot started successfully");
     } catch (err) {
-        log.error(err);
-        log.error("Failure during startup. Exiting");
-        process.exit(1);
+        if (err.code === 'TOKEN_INVALID' && adminMxid && !adminNotified) {
+            await notifyBridgeAdmin(client, adminMxid, `Your Discord bot token seems to be invalid, and the bridge cannot function. Please update it in your bridge settings and restart the bridge`);
```
Half-Shot: I think we should have the ability to customize this, e.g. linking to the correct page on EMS

src/discordas.ts
```diff
@@ -214,20 +216,67 @@ async function run(): Promise<void> {
...
+    [mxid: string]: string;
+}
+
+async function notifyBridgeAdmin(client: MatrixClient, adminMxid: string, message: string) {
+    log.verbose(`Looking up admin room for ${adminMxid}`);
+    let adminRooms = await client.getAccountData<AdminRooms>(ADMIN_ROOMS_KEY).catch(() => null) || {};
+    if (!adminRooms[adminMxid]) {
+        log.verbose(`Creating an admin room for ${adminMxid}`);
+        adminRooms[adminMxid] = await client.createRoom();
+        await client.inviteUser(adminMxid, adminRooms[adminMxid]);
```
Half-Shot: Any danger in inviting the users as part of the createRoom opts?

src/discordas.ts
```diff
@@ -214,20 +216,67 @@ async function run(): Promise<void> {
...
+        await client.setAccountData(ADMIN_ROOMS_KEY, adminRooms);
```
Half-Shot: Hah, clever. To avoid database usage?

src/discordas.ts
```diff
@@ -214,20 +216,67 @@ async function run(): Promise<void> {
...
     await appservice.begin();
     log.info(`Started listening on port ${port}`);
 
+}
+
+interface AdminRooms {
+    [mxid: string]: string;
+}
+
+async function notifyBridgeAdmin(client: MatrixClient, adminMxid: string, message: string) {
```
Half-Shot: Is there a chance we need to avoid this being called rapidly and creating lots of rooms?

Half-Shot (commented):
Initial thoughts

src/discordas.ts
tadzik: Yeah, I don't wanna add a whole new table for such a minor thing :)

src/discordas.ts
tadzik: Hmm, it will keep spamming rooms if it fails to invite the admin user (due to invalid MXID for example). Perhaps it should `try` it all and clean up the room if any of the steps after it fail.

src/discordas.ts
tadzik: No, I just haven't dug up the docs for that ;) Will give it a shot, might save me some work later on.

src/discordas.ts
tadzik: Should no longer do that as of 7b02c4f

src/discordas.ts
tadzik: Kept in as-is since it makes cleanup possible in case the invite didn't work for some reason.

src/discordas.ts
tadzik: In in a7adca9

src/discordas.ts
```diff
@@ -230,44 +228,70 @@ async function run(): Promise<void> {
...
+    await client.sendText(roomId, message)
 }
 
 let adminNotified = false;
 
 async function startDiscordBot(
-    discordbot: DiscordBot,
-    client: MatrixClient,
-    adminMxid: string|null,
+    discordbot:    DiscordBot,
```
Half-Shot: I do not like this indent :(

src/discordas.ts
```diff
@@ -214,20 +214,93 @@ async function run(): Promise<void> {
...
+    log.verbose(`Created ${roomId} to DM with ${mxid}`);
+    return roomId;
+}
+
+async function notifyBridgeAdmin(client: MatrixClient, adminMxid: string, message: string) {
+    const roomId = await ensureDMRoom(client, adminMxid);
+    await client.sendText(roomId, message)
+}
+
+let adminNotified = false;
```
justinbot: Seems odd to define a mutable variable here. Could it not be put inside `run`?

src/discordas.ts
tadzik: It breaks my heart, but I reverted it in b8492e5

src/discordas.ts
tadzik: Well, it's very specific to that startDiscordBot in particular. I'm guessing this whole logic could/should actually be moved to the DiscordBot itself, which already has a state so this extra flag in there wouldn't stand out so much. What do you think?

justinbot: That sounds reasonable. Having it in the global scope just made the flow more difficult to follow.

tadzik: Changed in d8c3be7.

The whole retry-loop is now available as `start()` (leaving existing APIs intact). `start()` is backed by a private `_start()` which does the actual work and keeps what used to be `adminNotified` in its parameter (which now acts as a state), now called `isRetry` which ends up being the same thing, more aptly named (imo).

Half-Shot (approved):
If tested, this looks good!

PR #747, opened 2021-11-22
Add Github CI for linting and tests

tadzik:
Both are currently [failing](https://github.com/tadzik/matrix-appservice-discord/actions/runs/1491202980) – but at least now we'll know :)


tadzik:
Suggestion: disable camelcase warning on the entire src/db?

Half-Shot (approved):
Seems reasonable!

Issue #748, opened 2021-11-26
Get 'Application service has not registered this user' error after successful setup


zaharzagrava:
First of all, I created discord application and discord bot (with Admin rights just in case), then I setup my config file:

``` yaml
bridge:
  # Domain part of the bridge, e.g. matrix.org
  domain: "d03f-37-139-160-239.ngrok.io"
  # This should be your publicly-facing URL because Discord may use it to
  # fetch media from the media store.
  homeserverUrl: "http://d03f-37-139-160-239.ngrok.io"
  # The TCP port on which the appservice runs on.
  port: 9005
  # Interval at which to process users in the 'presence queue'. If you have
  # 5 users, one user will be processed every 500 milliseconds according to the
  # value below. This has a minimum value of 250.
  # WARNING: This has a high chance of spamming the homeserver with presence
  # updates since it will send one each time somebody changes state or is online.
  presenceInterval: 500
  # Disable setting presence for 'ghost users' which means Discord users on Matrix
  # will not be shown as away or online.
  disablePresence: false
  # Disable sending typing notifications when somebody on Discord types.
  disableTypingNotifications: false
  # Disable deleting messages on Discord if a message is redacted on Matrix.
  disableDeletionForwarding: false
  # Disable portal bridging, where Matrix users can search for unbridged Discord
  # rooms on their Matrix server.
  disablePortalBridging: false
  # Enable users to bridge rooms using !discord commands. See
  # https://t2bot.io/discord for instructions.
  enableSelfServiceBridging: false
  # Disable sending of read receipts for Matrix events which have been
  # successfully bridged to Discord.
  disableReadReceipts: false
  # Disable Join Leave echos from matrix
  disableJoinLeaveNotifications: false
  # Disable Invite echos from matrix
  disableInviteNotifications: false
  # Auto-determine the language of code blocks (this can be CPU-intensive)
  determineCodeLanguage: false
# Authentication configuration for the discord bot.
auth:
  # This MUST be a string (wrapped in quotes)
  clientID: "913697282707308565"
  botToken: "I_HAVE_PASTED_MY_BOT_TOKEN_HERE"
  # You must enable "Privileged Gateway Intents" in your bot settings on discord.com (e.g. https://discord.com/developers/applications/913697282707308565/bot)
  # for this to work
  usePrivilegedIntents: true
logging:
  # What level should the logger output to the console at.
  console: "verbose" #silly, verbose, info, http, warn, error, silent
[...]

Issue #721, opened 2021-07-25
Error when using Docker container: 'User ID already taken'


hwittenborn:
I'm seeing this error, which appears to be causing everything to not work properly:

```
[bot-sdkMatrixLiteClient (REQ-1)] error: [ { errcode: 'M_USER_IN_USE', error: 'User ID already taken.' } ]
```

After this the container appears to just be quitting, and the bot doesn't do anything like join rooms.

d-513:
same error

Sorunome:
Those should be safe to ignore. What are your "not working" symptoms?

hwittenborn:
"Not working" is the following:

- The bot isn't joining
- The bot appeared to quit after the message in the first post showed up

I'm not sure why, but attempting to run `docker logs` on the container is also mysteriously not showing anything now.

hwittenborn:
Just turned on `verbose` logging, and it's outputting logs just fine again.

It's also saying that the bot started successfully now (don't know why the error's just gone now; I did change over to the `develop` image beforehand though), but the issue about the bot not joining rooms is still present.

hwittenborn:
Went back to the `latest` image, but I'm still not getting any dice with it joining rooms (the original error's still gone though, so it's a step in the right direction anyway).

Synkstar:
Im getting this same issue actually. Its just the bot doesnt join the matrix room when its invited but it shows up on discord as being online.

`discordrelay_1  | Dec-7 13:15:54.636 [bot-sdkMatrixLiteClient (REQ-1)] error: [ { errcode: 'M_USER_IN_USE', error: 'User ID already taken.' } ]
discordrelay_1  | PayloadTooLargeError: request entity too large
discordrelay_1  |     at readStream (/node_modules/raw-body/index.js:155:17)
discordrelay_1  |     at getRawBody (/node_modules/raw-body/index.js:108:12)
discordrelay_1  |     at read (/node_modules/body-parser/lib/read.js:77:3)
discordrelay_1  |     at jsonParser (/node_modules/body-parser/lib/types/json.js:135:5)
discordrelay_1  |     at Layer.handle [as handle_request] (/node_modules/matrix-bot-sdk/node_modules/express/lib/router/layer.js:95:5)
discordrelay_1  |     at trim_prefix (/node_modules/matrix-bot-sdk/node_modules/express/lib/router/index.js:317:13)
[...]

PR #751, opened 2021-12-04
Dockerfile build fails: move to python3

DeveloperTK:
As python 2 [was deprecated](https://www.python.org/doc/sunset-python-2/) quite some time ago and the package `python` removed from the alpine linux package registries, I suggest changing this to `py-pip`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 FROM node:14-alpine AS BUILD
 COPY . /tmp/src
 # install some dependencies needed for the build process
-RUN apk add --no-cache -t build-deps make gcc g++ python ca-certificates libc-dev wget git
+RUN apk add --no-cache -t build-deps make gcc g++ py-pip ca-certificates libc-dev wget git
 RUN cd /tmp/src \
     && yarn
 


auscompgeek:
Is python actually required?

Synkstar:
Yeah it is required I get an error about python not existing when I try to build the image.
`
Sending build context to Docker daemon  827.9kB
Step 1/13 : FROM node:14-alpine AS BUILD
 ---> 590973483ea0
Step 2/13 : COPY . /tmp/src
 ---> Using cache
 ---> fc427ae08ec9
Step 3/13 : RUN apk add --no-cache -t build-deps make gcc g++ python ca-certificates libc-dev wget git
 ---> Running in 9c6f8a8e4924
fetch https://dl-cdn.alpinelinux.org/alpine/v3.14/main/x86_64/APKINDEX.tar.gz
fetch https://dl-cdn.alpinelinux.org/alpine/v3.14/community/x86_64/APKINDEX.tar.gz
ERROR: unable to select packages:
  python (no such package):
    required by: build-deps-20211207.135802[python]
The command '/bin/sh -c apk add --no-cache -t build-deps make gcc g++ python ca-certificates libc-dev wget git' returned a non-zero code: 2
`
I'm also getting other problems when trying to run it from the images on docker hub. On the develop tag I get #748. But on the latest tag I get #721. I just used the changes from this pull and it built the container but I still get the same issues.

MightyCreak:
It's weird that `python` can't be installed, but [`py-pip`](https://pkgs.alpinelinux.org/package/v3.3/main/x86/py-pip), which depends on `python`, can.

Wouldn't it be better to migrate to python3?

Jaykul:
It would make more sense to me to move to python3

Miepee:
PR #798 is basically this PR but with `python3` instead of `py-pip`. Should this PR be closed then?

jaller94:
Superseded by #825

PR #771, opened 2022-01-29
Fixed where embeds on Discord could have HTML entities in their body

nicholemattera:
Fixed a bug where HTML entities would show up inside embeds on Discord, noticed it specifically when replying to someone. Also removed `escape-html` as it was no longer being used anywhere and `html-entities` (which was added) does both encoding and decoding.

Before:
image

After:
image


nicholemattera:
FYI: Found an issue in this where you can send `&#x40;everyone` in Matrix and it will ping everyone. Looking into how to fix that.

nicholemattera:
I have a PR with `matrix-discord-parser` (https://github.com/matrix-discord/matrix-discord-parser/pull/23), which should fix the apostrophe issue. Once that merges I'll create a new PR here to update the package.

PR #777, opened 2022-03-25
package.json: replace defunct git:// URLs

pacien:
Using Git over HTTPS instead.

Unencrypted `git://` does not seem to be supported by GitHub anymore:
https://github.blog/2021-09-01-improving-git-protocol-security-github/

This prevented the package from building on NixOS:
https://github.com/NixOS/nixpkgs/issues/165246

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
   },
   "homepage": "https://github.com/Half-Shot/matrix-appservice-discord#readme",
   "dependencies": {
-    "better-discord.js": "git://github.com/Sorunome/better-discord.js.git#b5a28499899fe2d9e6aa1aa3b3c5d693ae672117",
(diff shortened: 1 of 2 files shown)


AndrewFerr (changes requested):
Thanks for this!

There's still one more URL that needs updating, though: `discord-markdown` in `yarn.lock`.  I confirmed that `yarn install` times out with it left as-is (after having cleared `yarn`'s cache and `node_modules`).

This is also the reason why CI is failing.

AndrewFerr:
Actually, I opened a new PR that incorporates this one: https://github.com/matrix-org/matrix-appservice-discord/pull/790

If that one gets merged, this one can be discarded.

Issue #770, opened 2022-01-18
Thumbnail attachments


cardinalfan1:
Image attachments seemed to get processed fine but thumbnails are not showing up. I tried using a regular webhook and a slack webhook to discord and neither worked. Not sure how to troubleshoot

Issue #743, opened 2021-10-26
Can we provide an option to select the avatar image format?


Jaykul:
In order to solve problems with clients (or other networks) that don't support webp [(specifically, slack)](https://github.com/matrix-org/matrix-appservice-slack/issues/578), would it be possible to have the discord bridge fetch the .png images instead of [converting them on the mediarepo](https://github.com/matrix-org/matrix-doc/issues/3164)?

Discord provides the avatars in WEBP, JPG, PNG, and sometimes even GIF formats [per their documentation](https://discord.com/developers/docs/reference#image-formatting-cdn-endpoints), so it seems like a simple workaround would be to have _this bridge_ allow forcing the images to be PNG, for instance.

It's not really a perfect solution, but it seems like it would be better than what we have now.

jaller94:
I think the receiving bridge (Slack in this case) should convert the image as it knows what file formats are supported by the network it's bridging to.

Jaykul:
Sure, the Slack bridge could _convert_ the image, but Discord **already has** all the formats... 

In fact, the discordjs call `user.avatarUrl()` _takes_ [image URL options](https://discord.js.org/#/docs/discord.js/stable/typedef/StaticImageURLOptions) like `avatarURL({ format: 'png' })` which is why I'm suggesting _this_ bridge could just expose that as an option.

PR #792, opened 2022-06-07
Force Discord avatars to be PNGs in a friendly matter

### LINKED ISSUE #788: Profile pictures of Discord users not showing up on Matrix
**Describe the bug**
Profile pictures of Discord users do not show up on Matrix, but instead a default Matrix profile picture/avatar.

**To Reproduce**
1. Have a Discord Channel bridged to a Matrix room
2. Have a user in the Discord channel change their profile picture.
3. See error

**Expected behavior**
The Discord puppet should use the new pfp from the Discord user.

**Screenshots**
The event for the avatar change is send:
image
However, this is the actual image of the user:
image

**Additional context**
This was reproduced on the t2bot fork, so there is a small chance it may not happen on upstream here.

Miepee:
This can be considered a hack in order to fix #788 

It was discussed a bit in `#help:t2bot.io` on what the root cause of the issue might be, and it was found out that the bridge creates webp thumbnails for users on t2bot.io, which then often (not always for some reason) aren't mirrored to matrix.org (and potentially other HS), causing Clients to just display a default/broken avatar.

This PR circumvents the root cause, by just forcing all avatars to be in PNG format, as then thumbnails get mirrored correctly.
Only users who *update* their avatar will create revalidation. This is done in order to not revalidate *every user currently existing on the bridge* which would cause *quite* a bit of network traffic.

An alternative solution would be to implement #743; have a config file with a wished image file format, which is then used for checking the discord profile picture. I however am not that proficient with JS nor can I self-host/test my changes, hence this primitive solution.
Even better would be to figure out why the thumbnails aren't generates/mirrored in the first place. Possibly is related to #770, but not sure.

Either way, I'm opening this PR so that people who are self-hosting and are encountering the issue have a solution they can apply until the underlying root causes can be found and properly fixed.

### DIFF
--- a/src/matrixeventprocessor.ts
+++ b/src/matrixeventprocessor.ts
@@ -499,13 +499,13 @@ export class MatrixEventProcessor {
             if (userOrMember instanceof Discord.User) {
                 embed.setAuthor(
                     userOrMember.username,
-                    userOrMember.avatarURL() || undefined,
+                    userOrMember.avatarURL({ format: 'png' }) || undefined,
                 );
                 return;
             } else if (userOrMember instanceof Discord.GuildMember) {
                 embed.setAuthor(
                     userOrMember.displayName,
-                    userOrMember.user.avatarURL() || undefined,
+                    userOrMember.user.avatarURL({ format: 'png' }) || undefined,
                 );
                 return;
             }
--- a/src/usersyncroniser.ts
+++ b/src/usersyncroniser.ts
@@ -257,7 +257,7 @@ export class UserSyncroniser {
             userState.createUser = true;
             userState.displayName = displayName;
             if (discordUser.avatar) {
-                userState.avatarUrl = discordUser.avatarURL();
+                userState.avatarUrl = discordUser.avatarURL({ format: 'png' });
                 userState.avatarId = discordUser.avatar;
             }
             return userState;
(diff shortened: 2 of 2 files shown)


AndrewFerr:
WebP thumbnails work, but requires installing an extra library.  On Debian, it's `libwebp-dev`, and it fixed the problem for me on my own instance of this bridge.  I can't recall if it was the bridge or Synapse that was failing without the library, though.

If that still works, we can just document that library as an optional dependency & make PNG-only thumbnails a configuration setting.

And IIRC Discord sometimes sends image attachments as WebP too, not just avatars.

Miepee:
> And IIRC Discord sometimes sends image attachments as WebP too, not just avatars.

Unless there's something new I'm not aware of, image attachments are always send as-is, so it's dependent on what the user sends, not Discord. And thumbnails are just generated via passing `width` and `height` parameters to the attachment URL.

Regarding `libwebp` am not sure about that one. The bridge *did* create webp thumbnails on the HS it was hosted on (in this case t2bot.io), they just weren't federated to other homeservers (I.e. the [t2bot thumbnail exists](https://t2bot.io/_matrix/media/v3/thumbnail/t2bot.io/5a7cadbfc4f4b196ba83e4c43608a52dfe8dd91a?width=32&height=32&method=scale), but [matrix.org thumbnail does not](https://matrix.org/_matrix/media/v3/thumbnail/t2bot.io/5a7cadbfc4f4b196ba83e4c43608a52dfe8dd91a?width=32&height=32&method=scale)). 
Is that what the dependency is needed for, or would it be needed to create the thumbnails and upload it to the HS the bot is hosted on in the first place?

AndrewFerr:
> Unless there's something new I'm not aware of, image attachments are always send as-is, so it's dependent on what the user sends, not Discord.

Ah, that makes sense.  Though that suggests the bridge/homeservers ought to support WebP, lest images fail to be bridged.

> The bridge _did_ create webp thumbnails on the HS it was hosted on (in this case t2bot.io), they just weren't federated to other homeservers

It might be that Synapse is what requires the library, not the bridge.  That would explain federation issues like what you've described (i.e. any homeserver that doesn't have the library will fail).  I'll try to reproduce the problem to refresh my memory on what the actual error is.

Miepee:
Also worth noting that this issue only occurred "recently". It used to federate them just fine a month or so ago, and then it just suddenly... stopped. Which still makes me question `libwebp`s role in it, since from my experience at least, people don't just remove dependencies for a running homeserver.

Damaj301damaj-lol:
Seems like the matrix.org homeserver disabled webp support, which considering it is the biggest homeserver, it cannot be forgiven, so using the patch works

PR #796, opened 2022-06-16
Use `yarn` instead of `npm` in docs and scripts

AndrewFerr:
This is on top of https://github.com/matrix-org/matrix-appservice-discord/pull/794, which should be merged first.


CONTRIBUTING.md
```diff
@@ -23,6 +23,8 @@ inside the `MatrixEventProcessor` class.
 
 * You will need to [setup the bridge](https://github.com/Half-Shot/matrix-appservice-discord/tree/develop#setup-the-bridge) similarly to how we describe,
   but you should setup a homeserver locally on your development machine. We would recommend [Synapse](https://github.com/matrix-org/synapse).
+* We recommend using `yarn` for dependency management (v1 / Classic for the time being).
```
tadzik: TIL that there are newer versions. Are there any known incompatibilities that could bite us/users?

auscompgeek: Yarn Berry installs in a completely different incompatible layout to Yarn Classic.

auscompgeek: Anyway, probably be a good idea to enforce node and yarn versions in package.json so users don't get bitten.

AndrewFerr:
Enforcing Yarn was a lot more involved than I expected!

This uses the recommendation from https://github.com/yarnpkg/yarn/issues/4895#issuecomment-545644733 to tweak `package.json` to forbid usage of `npm` and set a requirement on which `yarn` version is used.

Since that doesn't actually install the required Yarn version, I added a package script to call handle that. It calls [`yarn set version`](https://yarnpkg.com/cli/set/version) (which appears to be supported by Yarn 1.x despite being documented only in later versions) to download & lock the latest 1.x release.

Note that [the recommended solution from the Yarn team](https://github.com/yarnpkg/berry/discussions/3064) is to lock versions with Corepack, or to keep the Yarn binary in version control.  Since Corepack is currently opt-in and checking in a script is cumbersome, I settled on using the `package.json` tweak.

package.json
```diff
@@ -3,14 +3,20 @@
   "version": "1.0.0",
   "description": "A bridge between Matrix and Discord",
   "main": "discordas.js",
+  "engines": {
+    "npm": "please-use-yarn",
+    "yarn": "^1.22.0",
+    "node": "12.x - 16.x"
+  },
```
Half-Shot, line 9: TIL this is valid syntax :o

package.json
```diff
@@ -3,14 +3,20 @@
...
   "scripts": {
+    "getyarn": "yarn set version 1.22",
```
Half-Shot: I'm ever so worried with locking yarn versions, because we will only forget to deal with this later. What's the rationale for minor version locking?

Half-Shot (changes requested):
Looks okay, I'm very worried about locking out r package manager versions and forgetting it. Would rather we just failed to run on yarn 2+ entirely.

package.json
AndrewFerr: Rationale is to prevent breakage by a later version of Yarn 1, but given that v1 is "stable" now, such breakage shouldn't be a threat.  Locking on any 1.x should be fine; what really matters is to prevent using Yarn >=2.

package.json
AndrewFerr: Which part, the `.x` or the `please`? :laughing:

package.json
AndrewFerr: As it turns out, Yarn 2 doesn't support `engines` at all!  Specifying a 1.x Yarn version in `engines` is thus useless because any v1 will satisfy it and any v2+ will ignore it (and `npm` doesn't detect check `yarn` at all).

Looks like a preinstall script is the most reliable option.

AndrewFerr: Never mind: while Yarn 1 runs preinstall scripts before resolving packages, Yarn 2+ runs them _after_ it, which I've verified experimentally & [is documented](https://yarnpkg.com/advanced/lifecycle-scripts):

> **postinstall** is called after a package's dependency tree changes are written to the disk...**preinstall** and **install** are called as part of **postinstall**.

See also:
- https://github.com/yarnpkg/berry/issues/2689
- https://github.com/yarnpkg/berry/issues/2701

As such, there doesn't seem to be a way to make `yarn` run a version-check script (or any script) before/at the start of the install step in Yarn 2+.

This means my running my `getyarn` script via `yarn getyarn` won't help either, because Yarn 2+ can't run it until after a successful `yarn`/`yarn install`--which fails on this project, thus defeating the purpose of the script.

Yarn does support enforcing a version via the `"packageManager"` key in `package.json`...but it's supported only Yarn 2+.

So [checking in the Yarn version to use](https://github.com/yarnpkg/berry/discussions/3064) seems to be the only reliable way to enforce Yarn 1 after all (and [`yarn policies set-version`](https://classic.yarnpkg.com/lang/en/docs/cli/policies/) exists to help set that up).  With that said, checking in a version is even more strict than just locking which version to download...

At this point, I'm inclined to give up programmatically enforcing Yarn 1 & just recommend using `yarn set version classic` as a manual setup step.  If/when we decide to move to Yarn 2+, these problems will be a thing of the past.

package.json
Half-Shot: The `.x`, I suspect the former might be a hack :p

package.json
Half-Shot: Yeah, I think there is significant value in just mentioning this in the README, which people can be referred to if they get stuck, imo.

Half-Shot (approved):
I am good with this change

PR #798, opened 2022-06-22
update `python` to `python3`

eniraa:
This should allow building Docker images for ARM, since the `python` package only exists for x86.

### DIFF
--- a/changelog.d/798.docker
+++ b/changelog.d/798.docker
@@ -1,1 +1,1 @@
-Update Dockerfile to allow building for ARM arches.
+Update Dockerfile to specifically use Python 3.
--- a/changelog.d/798.docker
+++ b/changelog.d/798.docker
@@ -1,0 +1,1 @@
+Update Dockerfile to allow building for ARM arches.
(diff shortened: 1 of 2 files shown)


changelog.d/798.docker
```diff
@@ -0,0 +1 @@
+Update Dockerfile to allow building for ARM arches.
```
Half-Shot: I see why you've done this, but I worry this might lead people to think we build ARM images too..which we don't yet. Could you maybe word it to sound like we've unstuck a dependency?

Half-Shot (commented):
Change looks good, just needs a changelog edit. And you need to [sign off](https://matrix-org.github.io/synapse/latest/development/contributing_guide.html#sign-off) :)

changelog.d/798.docker
eniraa: Done

changelog.d/798.docker
```diff
@@ -0,0 +1 @@
+Update Dockerfile to specifically use Python 3.
```
Half-Shot, line 1: We don't have a `docker` entry in our pyproject.toml file which is why this is failing. I'm happy to either have this be a `misc`, or you can add a Docker section yourself.

jaller94:
Superseded by https://github.com/matrix-org/matrix-appservice-discord/pull/825

PR #811, opened 2022-07-15
Drop support for Node 12, default to Node 16

Half-Shot:
Like our other matrix.org bridges, we are dropping support for Node 12. Node 12 is out of support, and should not be used https://nodejs.org/en/about/releases/

Node 18 is new and fresh, and we should support it.

Node 16 will become our new default as the current latest LTS.


AndrewFerr (approved):
Changes look simple & safe, and CI is happy, so I am too :slightly_smiling_face:

Issue #822, opened 2022-08-05
Prebuilt Docker images?


spantaleev:
Looks like v2.0.0 is not published on Docker Hub at [halfshot/matrix-appservice-discord](https://hub.docker.com/r/halfshot/matrix-appservice-discord), or at least not yet.

[t2bot/matrix-appservice-discord](https://hub.docker.com/r/t2bot/matrix-appservice-discord) is more up-to-date, but only publishes a `:latest` tag.

Do you plan on providing up-to-date prebuilt/published Docker images or are people expected to build images themselves?

turt2live:
Just in case: **Do not use the t2bot image in production - it is unstable and not supported for anyone other than t2bot.io**

TheNamelessWonderer:
The docker image won't build until https://github.com/matrix-org/matrix-appservice-discord/pull/798 is merged.

spantaleev:
That PR looks abandoned, unfortunately.

Half-Shot:
As of today, we are now publishing against ghcr.io https://github.com/matrix-org/matrix-appservice-discord/pkgs/container/matrix-appservice-discord/33898277?tag=develop. New release is expected to land there soon!

PR #825, opened 2022-08-08
Update yarn.lock and fix Docker builds

### LINKED ISSUE #824: `yarn.lock` seems out of sync for v2.0.0
Installation of release v2.0.0 using the strict version requirements from the
lock file fails:

```log
❯ nix log /nix/store/414s3b1p2krfsnn9hhv9yk08xkxw390c-matrix-appservice-discord-modules-2.0.0.drv
@nix { "action": "setPhase", "phase": "patchPhase" }
patching sources
@nix { "action": "setPhase", "phase": "configurePhase" }
configuring
@nix { "action": "setPhase", "phase": "buildPhase" }
building
yarn config v1.22.19
success Set "yarn-offline-mirror" to "/nix/store/gxmyj7wx1vzaifqjsjby3kdmfqnklni2-offline".
Done in 0.02s.
yarn install v1.22.19
[1/4] Resolving packages...
error Couldn't find any versions for "@types/node" that matches "^14" in our cache (possible versions are ""). This is usually caused by a missing entry in the lockfile, running Yarn without the --offline flag may help fix this issue.
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
```

This is problematic for some distribution packages like on nixpkgs.

It seems to be due to `yarn.lock` being out of sync with `package.json`.
It could be a good idea to check that through the CI, in particular for tagged
releases.

jaller94:
* Fixes #824 
* Part of #822 – given that @Half-Shot publishes the Docker image.
* Supersedes #798
* Upgrades `matrix-appservice-bridge` to v5.0.0 which requires Node.js 16, 17 or 18

### DIFF
--- a/changelog.d/825.misc
+++ b/changelog.d/825.misc
@@ -1,1 +1,0 @@
-Require Node.js 16, 17 or 18 to run. Update yarn.lock and fix Docker builds.
--- a/changelog.d/825.misc
+++ b/changelog.d/825.misc
@@ -1,0 +1,1 @@
+Update yarn.lock and fix Docker builds
--- a/changelog.d/825.misc
+++ b/changelog.d/825.misc
@@ -1,1 +1,1 @@
-Update yarn.lock and fix Docker builds
+Update yarn.lock and fix Docker builds.
--- a/changelog.d/825.misc
+++ b/changelog.d/825.misc
@@ -1,1 +1,1 @@
-Update yarn.lock and fix Docker builds.
+Require Node.js 16, 17 or 18 to run. Update yarn.lock and fix Docker builds.
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -11,6 +11,8 @@ jobs:
       uses: actions/setup-node@v1
       with:
         node-version: 16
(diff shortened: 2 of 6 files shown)


jaller94:
With [3e491ca](https://github.com/matrix-org/matrix-appservice-discord/pull/825/commits/3e491ca7b3aed2e58a14c18027a4f14bbc29658a) this is now a breaking change that requires a new major version.

Because of the library `matrix-appservice-bridge` v5.0.0 the supported Node.js versions are 16 and 18.

changelog.d/825.misc
```diff
@@ -0,0 +1 @@
+Require Node.js 16, 17 or 18 to run. Update yarn.lock and fix Docker builds.
```
Half-Shot:
```suggestion
Minimum required Node.js version is now 16.
```

Half-Shot: This also needs to be a `.removal` to indicate that it's a breaking change.

Half-Shot (approved):
I don't think we need to be explicit about all the versions we support, 16 is the headline. Otherwise, lgtm.

PR #826, opened 2022-08-08
Add workflow to build container image

Half-Shot:
Partial fix for #822 

The docker image build process used to be handled by dockerhub, but a while back they stopped doing builds for free (I believe?) and we never really fixed the issue.

Rather than going through the hoops to fix dockerhub, for the next release we can just direct users to use `ghcr.io/matrix-org/matrix-appservice-discord`.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,0 +1,44 @@
+name: Container Image
+
+on:
+  push:
+    branches: [ main ]
+    tags: [ 'v*' ]
+  pull_request:
+    branches: [ main ]
+
+env:
+  IMAGE_NAME: ${{ github.repository }}
+  REGISTRY: ghcr.io
+
+jobs:
+  push:
+    runs-on: ubuntu-latest
+
+    permissions:
+      contents: read
+      packages: write
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      - name: Log into registry ${{ env.REGISTRY }}
+        uses: docker/login-action@v1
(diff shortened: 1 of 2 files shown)


.github/workflows/docker.yml
```diff
@@ -0,0 +1,44 @@
...
+      - name: Extract Docker metadata
+        id: meta
+        uses: docker/metadata-action@v3
+        with:
+          images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
+
+      - name: Build and push Docker image
+        uses: docker/build-push-action@v2
+        with:
+          push: ${{ github.event_name != 'pull_request' }}
```
Half-Shot, line 42: Notably, this *won't* automatically make latest == develop, which is what we did previously. I don't know if we have a set standard across our other bridges.

Half-Shot: (matrix-hookshot *does* push `latest` when things hit `main`)

jaller94: I'd like us to automate the update of `latest` as well, though it may follow in a separate PR.

jaller94 (approved):
Looks good to me.

Half-Shot:
Depends on #825

.github/workflows/docker.yml
Half-Shot: Believe this will do the right thing, albeit it will publish `develop` as `develop` and `latest`/`v.X.X` for versioned releases. I'm quite comfortable with that, anyway.

Issue #772, opened 2022-02-09
You're invited to talk on Matrix


heinrich5991:
Happens when a Matrix message mentioning Matrix users gets bridged to Discord:

![Matrix message bridged to Discord with a "You're invited to talk on Matrix" embed](https://user-images.githubusercontent.com/366230/153195011-b9fbc250-8c2c-4caf-9ec0-d0a55082fad4.png)

In native Discord messages, you can enclose links in <> to avoid the embed, like
```
thanks a lot <https://matrix.to/#/@jupeyy:matrix.org> :)
```

Does something like that work here?

erkinalp:
On webhooks, the only option is to suppress *all* embeds with SUPPRESS_EMBEDS.

heinrich5991:
Sounds like a good idea at least on messages only containing links for mentions.

TurBoss:
hello, I'm unable to figure where i should set the "SUPPRESS_EMBEDS" flag

erkinalp:
@TurBoss set 4 in the `flags` field of the Discord-side webhook.

PR #829, opened 2022-08-11
Update to @mx-puppet/matrix-discord-parser 0.1.10

### LINKED ISSUE #652: strip out matrix.to hyperlinks from mentions
currently every Matrix mention causes a matrix.to link preview to unfurl in Discord.  This is partially due to https://github.com/matrix-org/matrix.to/issues/182, but presumably we should be mapping them to Discord mentions rather than leaking matrix.to URLs everywhere

### LINKED ISSUE #760: Cannot @mention Matrix users in Discord
While Discord mentions are correctly transformed into Matrix mentions, in Discord I can only @mention Discord users, not Matrix users. Makes sense since presence is not synced, but mentions (ie. sending notifications) are a very fundamental feature of any chat system so those not working degrades the user experience a lot. IMO some poor-man's mention support would be better than nothing - just check for any plaintext `@<matrix-nick>` string in Discord messages and convert to a mention if found.

jlu5:
This matrix-discord-parser update pulls in a few changes, including:

- https://gitlab.com/mx-puppet/discord/matrix-discord-parser/-/merge_requests/21, which removes matrix.to links for non-Discord user mentions. Fixes #652 and #772
- https://gitlab.com/mx-puppet/discord/matrix-discord-parser/-/commit/52592702afd3fb9e9305f04b785a0001e59f1ef1, which adds support for Discord -> Matrix mentions. Fixes #760

I've updated the tests to match these new behaviours.

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,1 +1,2 @@
 engine-strict = true
+@mx-puppet:registry="https://gitlab.com/api/v4/packages/npm/"
--- a/changelog.d/829.bugfix
+++ b/changelog.d/829.bugfix
@@ -1,0 +1,2 @@
+Remove matrix.to hyperlinks when relaying non-Discord user mentions to Discord.
+Fix mentioning Matrix users in Discord.
--- a/package.json
+++ b/package.json
@@ -40,6 +40,7 @@
   },
   "homepage": "https://github.com/Half-Shot/matrix-appservice-discord#readme",
   "dependencies": {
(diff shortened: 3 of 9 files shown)


Half-Shot (changes requested):
Changes look sound, you are one changelog file away from a merge!

For the changelog, I'd just identify what this changes for the end user (removing matrix.to links etc)

Half-Shot (approved):

PR #836, opened 2022-08-24
Add option to disable room topic echos from Matrix

Miepee:
This PR adds another config value, in order to disable forwarding room topic echos on Matrix to Discord.

### DIFF
--- a/changelog.d/836.feature
+++ b/changelog.d/836.feature
@@ -1,0 +1,1 @@
+Adds a config value, in order to disable forwarding room topic changes from Matrix to Discord (`disableRoomTopicNotifications`, false by default).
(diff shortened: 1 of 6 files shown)


Half-Shot (changes requested):
Changes look good, but can you add a test next to https://github.com/matrix-org/matrix-appservice-discord/blob/develop/test/test_matrixeventprocessor.ts#L345-L356? I think all you need to do is modify the createMatrixEventProcessor function to take a parameter which customises the config, and then test for no topic notifs.

Thanks!

Miepee:
I originally didn't add a test case, because the join/leave and invite echos don't have a test case either for not sending a message when the config has it disabled.
Should I add the test cases for those too then?

Half-Shot:
> I originally didn't add a test case, because the join/leave and invite echos don't have a test case either for not sending a message when the config has it disabled. Should I add the test cases for those too then?

That's definitely an oversight. Extra points if you do those too, but I'm also happy just to accept changes for the things you tested.

Half-Shot (approved):
Looks good to me!

PR #855, opened 2022-09-20
chore: update readme, contributing.md, and fix typos

SethFalco:
* Add an [EditorConfig](https://editorconfig.org/) file so it's easier code editor settings don't get in the way while contributing to the project.
* Fix typos in CONTRIBUTING.md, also the link fragment in one of the URLs was outdated. (setup vs set-up)
* Docker gives warning about apt vs apt-get, could we just make this user apt-get so we don't have to deal with them? ^-^' (Plus I guess it's better to go the stable API route anyway.)
* Very minor code change, just removes a redundant import and makes nouns like Discord/Matrix match the proper capitalization.
* While setting up the bridge, I added the 2 config properties from the README, but then the bot failed to start becasue I had 2 remove those 2 config properties. If they're outdated, I guess they shouldn't be suggested on the README either?
* Update broken link in comments.

---

These are all just small things I found while working on https://github.com/matrix-org/matrix-appservice-discord/pull/854. I didn't do a CHANGELOG entry intentionally since these are just chores, let me know if you want me to add one anyway.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,20 @@
+root = true
+
+[*]
+indent_style = space
+indent_size = 4
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+curly_bracket_next_line = false
+spaces_around_operators = true
+
+[*.{js,ts}]
+quote_type = double
+
+[*.{markdown,md}]
+trim_trailing_whitespace = false
+
+[*.{yaml,yml}]
+indent_size = 2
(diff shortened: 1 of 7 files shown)


Miepee:
> Docker gives warning about apt vs apt-get

Why does Docker give warnings? AFAIK `apt-get` is deprecated, so programs should use `apt` whenever possible

SethFalco:
apt-get isn't deprecated! 
In scripts, `apt` will throw a warning because it's not considered a stable interface.

So for scripts/Docker, it's recommended to use `apt-get`.

https://askubuntu.com/a/829869/938482

PR #858, opened 2022-10-07
Track a slightly fresher fork of better-discord.js

tadzik:
This fixes a crash related to voice channel message handling.

### DIFF
--- a/changelog.d/858.bugfix
+++ b/changelog.d/858.bugfix
@@ -1,0 +1,1 @@
+Prevent crashes when handling messages sent to voice channels.
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
   "homepage": "https://github.com/Half-Shot/matrix-appservice-discord#readme",
   "dependencies": {
     "@mx-puppet/matrix-discord-parser": "0.1.10",
-    "better-discord.js": "git+https://github.com/Sorunome/better-discord.js.git#b5a28499899fe2d9e6aa1aa3b3c5d693ae672117",
+    "better-discord.js": "github:matrix-org/better-discord.js#5024781db755259e88abe915630b7d5b3ba5f48f",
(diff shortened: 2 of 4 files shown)


Half-Shot (approved):
Seems good, changelog please :)

FSG-Cat:
If I may ask why are we switching to a branch that is a a single commit ahead? Isn't it a better idea to then climb up the tree to discordjs main if we want a more updated version?

Half-Shot:
Hiya @FSG-Cat, it's a good point:

We're updating to fix a specific bug, and are considering updating to stock in the future. It would take considerable effort to refactor the bridge for the moment, so we want to ensure we fix the more pressing bug, then circle back to refactoring later :)

Thanks for raising it!

Miepee:
Which bug was it? If it's user facing, maybe mention that in the changelog as well?

FSG-Cat:
> Hiya @FSG-Cat, it's a good point:
> 
> We're updating to fix a specific bug, and are considering updating to stock in the future. It would take considerable effort to refactor the bridge for the moment, so we want to ensure we fix the more pressing bug, then circle back to refactoring later :)
> 
> Thanks for raising it!

Thanks for the information.

turt2live:
uh, you should be able to switch to upstream.

erkinalp:
alternative: switch to fosscord.js which is a fork of discord.js with third party instance support.

tadzik:
I clarified the issue in both the commit message and the changelog.

The switch to upstream discord/fosscord.js is very much on our radar, but given how far behind we are and all the API changes that happened in the upstream since the version we're on, this is a significantly bigger effort that'll take us a while to complete.

This is a hotfix to fix a voice-channel related crash, just to improve the stability of the bridge while the proper upstream sync is still brewing.

PR #862, opened 2022-10-09
feat: add discord > matrix reaction support

### LINKED ISSUE #469: Send m.annotation events for message reactions
See https://github.com/matrix-org/matrix-doc/pull/1849 for details.

SethFalco:
### Related

* Closes https://github.com/matrix-org/matrix-appservice-discord/issues/469
* Closes https://github.com/matrix-org/matrix-appservice-discord/pull/784

Adds reaction support in one direction (Discord → Matrix).

Supports:
* Adding reactions
* Removing reactions
* Removing all reactions

There is an "edge-case" with some emotes.

Here is a thumbs-up emoji on Discord: 👍 (Base64: 8J+RjQo=)
Here is a thumbs-up emoji on Element: 👍️ (Base64: 8J+Rje+4jwo=)

If someone reacts with a thumbs-up from Discord, then someone clicks that reaction in Element (and presumably any Matrix client) it's handled perfectly fine.

However, if a user separately goes through the Element interface to do a thumbs-up emoji, because it's technically a different character, you'll see 2 separate reactions for a thumbs-up.

image
> When a Discord user does a thumbs-up, and an Element users manually performs a thumbs-up without clicking the pre-existing reaction.

### Notes

There are a few ways we can approach bridging reactions with custom emojis.

| | | |
|---|---|---
| name | But on Discord custom emotes can have the same name, so if the same user does multiple reactions with emojis with the same name, it'll only come through as one. If they do 2 reactions, then remove one, Discord will show the other, while Matrix will show neither.
| name:id | But on Discord, people can rename emotes. So if they react, then rename the emote, then remove the reaction, it will count as a different emote and won't be removed on the Matrix side.
| id | But on Matrix, this is basically a meaningless number, at that point we might as well not bridge custom emojis at all.

It looks like in future it should be supported to use external images as reactions via MXC URLs. I don't do this yet because the `shortcode` is not exposed in the SDK and it's not clearly documented behavior, but in future can update the bridge to support that.

Based on what I've seen, others use the format `:name:` for the `shortcode` property. Based on this, I've decided to use the same format already. That way when we're in a position to support MXC URLs and the `shortcode` property, we can just move the value of `key` to `shortcode` for those cases. 

### Screenshots

image
> A message and 3 reactions all sent from Discord.

image
> How the same message/reactions appear on Matrix when bridged.

### Notes

* `getRelationsForEvent` was moved out of `unstableApis` in the **main** branch for **matrix-bot-sdk**. Eventually, when we update to a version of **matrix-appservice-bridge** that uses that version we should switch it out stop using it from `unstableApis`.

### DIFF
--- a/src/bot.ts
+++ b/src/bot.ts
@@ -263,6 +263,21 @@ export class DiscordBot {
                 await this.channelSync.OnGuildDelete(guild);
             } catch (err) { log.error("Exception thrown while handling \"guildDelete\" event", err); }
         });
+        client.on("messageReactionAdd", async (reaction, user) => {
+            try {
+                await this.OnMessageReactionAdd(reaction, user);
+            } catch (err) { log.error("Exception thrown while handling \"messageReactionAdd\" event", err); }
+        });
+        client.on("messageReactionRemove", async (reaction, user) => {
+            try {
+                await this.OnMessageReactionRemove(reaction, user);
+            } catch (err) { log.error("Exception thrown while handling \"messageReactionRemove\" event", err); }
+        });
+        client.on("messageReactionRemoveAll", async (message) => {
+            try {
+                await this.OnMessageReactionRemoveAll(message);
+            } catch (err) { log.error("Exception thrown while handling \"messageReactionRemoveAll\" event", err); }
+        });
 
         // Due to messages often arriving before we get a response from the send call,
         // messages get delayed from discord. We use Util.DelayedPromise to handle this.
@@ -1175,6 +1190,143 @@ export class DiscordBot {
         }
         newMsg.content = `Edit: ${newMsg.content}`;
         await this.OnMessage(newMsg);
+    }
+
+    public async OnMessageReactionAdd(reaction: Discord.MessageReaction, user: Discord.User | Discord.PartialUser) {
+        const message = reaction.message;
+        log.info(`Got message reaction add event for ${message.id} with ${reaction.emoji.name}`);
+
+        let rooms: string[];
+
+        try {
+            rooms = await this.channelSync.GetRoomIdsFromChannel(message.channel);
+
+            if (rooms === null) {
+                throw Error();
+            }
+        } catch (err) {
+            log.verbose("No bridged rooms to send message to. Oh well.");
+            MetricPeg.get.requestOutcome(message.id, true, "dropped");
+            return;
+        }
+
+        const intent = this.GetIntentFromDiscordMember(user);
+        await intent.ensureRegistered();
+        this.userActivity.updateUserActivity(intent.userId);
+
+        const storeEvent = await this.store.Get(DbEvent, {
+            discord_id: message.id
+        });
+
+        if (!storeEvent?.Result) {
+            return;
+        }
+
+        while (storeEvent.Next()) {
+            const matrixIds = storeEvent.MatrixId.split(";");
+
+            for (const room of rooms) {
+                const reactionEventId = await intent.underlyingClient.unstableApis.addReactionToEvent(
+                    room,
+                    matrixIds[0],
+                    reaction.emoji.id ? `:${reaction.emoji.name}:` : reaction.emoji.name
+                );
+
+                const event = new DbEvent();
+                event.MatrixId = `${reactionEventId};${room}`;
+                event.DiscordId = message.id;
+                event.ChannelId = message.channel.id;
+                if (message.guild) {
+                    event.GuildId = message.guild.id;
+                }
+
+                await this.store.Insert(event);
+            }
+        }
+    }
+
+    public async OnMessageReactionRemove(reaction: Discord.MessageReaction, user: Discord.User | Discord.PartialUser) {
+        const message = reaction.message;
+        log.info(`Got message reaction remove event for ${message.id} with ${reaction.emoji.name}`);
+
+        const intent = this.GetIntentFromDiscordMember(user);
+        await intent.ensureRegistered();
+        this.userActivity.updateUserActivity(intent.userId);
+
+        const storeEvent = await this.store.Get(DbEvent, {
+            discord_id: message.id,
+        });
+
+        if (!storeEvent?.Result) {
+            return;
+        }
+
+        while (storeEvent.Next()) {
+            const [ eventId, roomId ] = storeEvent.MatrixId.split(";");
+            const underlyingClient = intent.underlyingClient;
+
+            const { chunk } = await underlyingClient.unstableApis.getRelationsForEvent(
+                roomId,
+                eventId,
+                "m.annotation"
+            );
+
+            const event = chunk.find((event) => {
+                if (event.sender !== intent.userId) {
+                    return false;
+                }
+
+                return event.content["m.relates_to"].key === reaction.emoji.name;
+            });
+
+            if (!event) {
+                return;
+            }
+
+            const { room_id, event_id } = event;
+
+            try {
+                await underlyingClient.redactEvent(room_id, event_id);
+            } catch (ex) {
+                log.warn(`Failed to delete ${storeEvent.DiscordId}, retrying as bot`);
+                try {
+                    await this.bridge.botIntent.underlyingClient.redactEvent(room_id, event_id);
+                } catch (ex) {
+                    log.warn(`Failed to delete ${event_id}, giving up`);
+                }
+            }
+        }
+    }
+
+    public async OnMessageReactionRemoveAll(message: Discord.Message | Discord.PartialMessage) {
+        log.info(`Got message reaction remove all event for ${message.id}`);
+
+        const storeEvent = await this.store.Get(DbEvent, {
+            discord_id: message.id,
+        });
+
+        if (!storeEvent?.Result) {
+            return;
+        }
+
+        while (storeEvent.Next()) {
+            const [ eventId, roomId ] = storeEvent.MatrixId.split(";");
+            const underlyingClient = this.bridge.botIntent.underlyingClient;
+
+            const { chunk } = await underlyingClient.unstableApis.getRelationsForEvent(
+                roomId,
+                eventId,
+                "m.annotation"
+            );
+
+            await Promise.all(chunk.map(async (event) => {
+                try {
+                    return await underlyingClient.redactEvent(event.room_id, event.event_id);
+                } catch (ex) {
+                    log.warn(`Failed to delete ${event.event_id}, giving up`);
+                }
+            }));
+        }
     }
 
     private async DeleteDiscordMessage(msg: Discord.Message) {
--- a/src/bot.ts
+++ b/src/bot.ts
@@ -1194,18 +1194,15 @@ export class DiscordBot {
 
     public async OnMessageReactionAdd(reaction: Discord.MessageReaction, user: Discord.User | Discord.PartialUser) {
         const message = reaction.message;
-        log.info(`Got message reaction add event for ${message.id} with ${reaction.emoji.name}`);
+        const reactionName = reaction.emoji.name;
+        log.verbose(`Got message reaction add event for ${message.id} with ${reactionName}`);
 
         let rooms: string[];
 
         try {
             rooms = await this.channelSync.GetRoomIdsFromChannel(message.channel);
-
-            if (rooms === null) {
-                throw Error();
-            }
         } catch (err) {
-            log.verbose("No bridged rooms to send message to. Oh well.");
+            log.verbose(`No bridged rooms to forward reaction to. Reaction Event: ${reaction}`);
             MetricPeg.get.requestOutcome(message.id, true, "dropped");
             return;
         }
@@ -1219,6 +1216,7 @@ export class DiscordBot {
         });
 
         if (!storeEvent?.Result) {
+            log.verbose(`Received add reaction event for untracked message. Dropping! Reaction Event: ${reaction}`);
             return;
         }
 
@@ -1229,7 +1227,7 @@ export class DiscordBot {
                 const reactionEventId = await intent.underlyingClient.unstableApis.addReactionToEvent(
                     room,
                     matrixIds[0],
-                    reaction.emoji.id ? `:${reaction.emoji.name}:` : reaction.emoji.name
+                    reaction.emoji.id ? `:${reactionName}:` : reactionName
                 );
 
                 const event = new DbEvent();
@@ -1247,7 +1245,7 @@ export class DiscordBot {
 
     public async OnMessageReactionRemove(reaction: Discord.MessageReaction, user: Discord.User | Discord.PartialUser) {
         const message = reaction.message;
-        log.info(`Got message reaction remove event for ${message.id} with ${reaction.emoji.name}`);
+        log.verbose(`Got message reaction remove event for ${message.id} with ${reaction.emoji.name}`);
 
         const intent = this.GetIntentFromDiscordMember(user);
         await intent.ensureRegistered();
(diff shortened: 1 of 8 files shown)


tezlm:
discord custom emojis can be done by setting a m.relates_to `key` to an `mxc://` url and content-level `shortcode` to the shortcode

```js
{
    "shortcode": "something",
    "m.relates_to": {
        "event_id": "$abc123",
        "key": "mxc://servername/someidhere",
        "rel_type": "m.annotation"
    }
}
```

SethFalco:
That's interesting. The MXC URL can be done already, but the SDK doesn't expose the `shortcode` property.

I'll leave this as-is for now, but I'll check what you said and verify if that's actually how it works. If so, I can open a PR to matrix-bot-sdk to include this, then come back to fix it here in a 2nd iteration if that's cool with maintainers.

Any chance you can link to where doing images as reactions is documented/defined?

D4ndellion:
I believe the closest thing to documentation on that is the MSC: https://github.com/matrix-org/matrix-spec-proposals/pull/3746

A bigger argument is probably that mx-puppet-discord, cinny, fluffychat, beeper, some element patches floating about, and probably more clients, all special case seeing an mxc url and render it as an image

src/bot.ts
```diff
@@ -1177,6 +1192,143 @@ export class DiscordBot {
         await this.OnMessage(newMsg);
     }
 
+    public async OnMessageReactionAdd(reaction: Discord.MessageReaction, user: Discord.User | Discord.PartialUser) {
+        const message = reaction.message;
+        log.info(`Got message reaction add event for ${message.id} with ${reaction.emoji.name}`);
```
tadzik: I'd make this `log.debug`, but I don't blame you for following the pattern after the other handlers.

src/bot.ts
```diff
@@ -1177,6 +1192,143 @@ export class DiscordBot {
...
+
+        let rooms: string[];
+
+        try {
+            rooms = await this.channelSync.GetRoomIdsFromChannel(message.channel);
+
+            if (rooms === null) {
```
tadzik: I don't think this will ever happen – `GetRoomIdsFromChannel` either throws (rejects), or returns a list.

src/bot.ts
```diff
@@ -1177,6 +1192,143 @@ export class DiscordBot {
...
+                throw Error();
+            }
+        } catch (err) {
+            log.verbose("No bridged rooms to send message to. Oh well.");
```
tadzik: I'd include the `message` here to potentially aid debugging this.

src/bot.ts
```diff
@@ -1177,6 +1192,143 @@ export class DiscordBot {
...
+        const intent = this.GetIntentFromDiscordMember(user);
+        await intent.ensureRegistered();
+        this.userActivity.updateUserActivity(intent.userId);
+
+        const storeEvent = await this.store.Get(DbEvent, {
+            discord_id: message.id
+        });
+
+        if (!storeEvent?.Result) {
+            return;
```
tadzik, line 1206: Maybe `log.verbose(dropping a reaction to a nonexistent message ${message})`? Just in case.

src/bot.ts
```diff
@@ -1177,6 +1192,143 @@ export class DiscordBot {
...
+        }
+
+        while (storeEvent.Next()) {
+            const matrixIds = storeEvent.MatrixId.split(";");
+
+            for (const room of rooms) {
```
tadzik: I'm wondering if we need `rooms` here at all (this loop and the lookup above).

The `storeEvent`s we process here already contain all the `${event}:${roomId}` information, so we should be able to just send the reaction straight to the relevant room from the `while (storeEvent.Next()) {` loop, like so:

```
while (storeEvent.Next()) {
     const [messageEventId, roomId] = storeEvent.MatrixId.split(";");
     const redactionEventId = await intent.underlyingClient.unstableApis.addReactionToEvent(
         roomId, messageEventId, reaction.emoji.id...
     );
} 
```

src/bot.ts
```diff
@@ -1177,6 +1192,143 @@ export class DiscordBot {
...
+                }
+
+                await this.store.Insert(event);
+            }
+        }
+    }
+
+    public async OnMessageReactionRemove(reaction: Discord.MessageReaction, user: Discord.User | Discord.PartialUser) {
+        const message = reaction.message;
+        log.info(`Got message reaction remove event for ${message.id} with ${reaction.emoji.name}`);
```
tadzik: `log.debug()`?

src/bot.ts
```diff
@@ -1177,6 +1192,143 @@ export class DiscordBot {
...
+        const intent = this.GetIntentFromDiscordMember(user);
+        await intent.ensureRegistered();
+        this.userActivity.updateUserActivity(intent.userId);
+
+        const storeEvent = await this.store.Get(DbEvent, {
+            discord_id: message.id,
+        });
+
+        if (!storeEvent?.Result) {
+            return;
```
tadzik, line 1252: `log.verbose()` here, as above?

src/bot.ts
```diff
@@ -1177,6 +1192,143 @@ export class DiscordBot {
...
+            const event = chunk.find((event) => {
+                if (event.sender !== intent.userId) {
+                    return false;
+                }
+
+                return event.content["m.relates_to"].key === reaction.emoji.name;
+            });
+
+            if (!event) {
+                return;
```
tadzik, line 1288: Let's add debug logs here as well (these should generally not happen, so we may find ourselves investigating these if they end up not working).

src/bot.ts
```diff
@@ -1177,6 +1192,143 @@ export class DiscordBot {
...
+
+            const { chunk } = await underlyingClient.unstableApis.getRelationsForEvent(
+                roomId,
+                eventId,
+                "m.annotation"
+            );
+
+            await Promise.all(chunk.map(async (event) => {
+                try {
+                    return await underlyingClient.redactEvent(event.room_id, event.event_id);
```
tadzik, line 1332: I'd check if the sender was us, to not send tons of requests that are destined to fail.

src/db/dbdataevent.ts
```diff
@@ -19,7 +19,9 @@ import { IDbDataMany } from "./dbdatainterface";
 import { ISqlCommandParameters } from "./connector";
 
 export class DbEvent implements IDbDataMany {
+    /** Matrix ID of event. */
```
tadzik: This is actually a `MatrixID:RoomID` string.

tadzik (changes requested):

src/bot.ts
SethFalco: Though is no `log.debug` exposed. Perhaps `log.verbose` works?

Edit: Assuming that's just a typo as I can see you say `verbose` in later comments.

src/bot.ts
SethFalco: I'm assuming since this is a `reaction` event, it'd be best to log `reaction` rather than `message`. Let me know if you disagree!

src/bot.ts
SethFalco: Rather than saying the message doesn't exist, maybe it's better to say it's untracked?

When the logs are separated from the code (which is how I'd expect it to be consumed quite often, especially for bug reports etc) readers may think it means the Discord message doesn't exist.

The term "untracked" makes clear it's just a message the bridge wasn't aware of, or perhaps wasn't online to the original message at the time it was sent.

i.e.

```ts
log.verbose(`Received add reaction event for untracked message. Dropping! Reaction Event: ${reaction}`);
```

src/bot.ts
SethFalco: This one might be my ignorance, but it looked to me in the code there is/was support for a channel to be bridged to multiple rooms, so I was thinking we'd have to check and iterate over each room the channel is bridges too. :thinking: 

Is that wrong?

SethFalco:
I believe I've fixed all comments except 1 which I've left unresolved.
Feel free to unresolve any of them if you disagree with how I handled it.

mangofeet:
I was testing this and it all seems to be working fine, but I'm getting this error

```
Exception thrown while handling "messageReactionRemove" event MatrixError: M_FORBIDDEN: User @_discord_<discord id>:example.com not in room !<room id>:example.com, and room previews are disabled
```

The room in question is a private room that has the bridge bot in it with admin access. Despite this error, everything appears to be functioning correctly.

SethFalco:
Hey @mangofeet, thanks for testing this!

Any chance you could provide more in-depth steps to reproduce or a stack trace?
If not, I'll get my environment up when I have time and try to check this out.

salixor:
Hi there ! I'm currently using this change on two servers with a few people. Thanks a lot for it !

I have two minor issues with it.

The first one, as @mangofeet reported, is that we are getting the same errors :

- `Exception thrown while handling "messageReactionAdd" event MatrixError: M_FORBIDDEN: User @_discord_<id>:<homeserver> not in room <room>`
- `Exception thrown while handling "messageReactionRemove" event MatrixError: M_FORBIDDEN: User @_discord_<id>:<homeserver> not in room <room>, and room previews are disabled`

A way to replicate the issue :

- bridge a channel
- have a Discord user which never posted in this channel since the briding react on a message
- have the same user remove their reaction (still without sending any message)
- the reaction is not bridged and the error is found in logs

Shouldn't the ghost be invited before trying to create a reaction on Matrix's side ?

-------------------------

The second issue has to do with Discord bot users. It seems their reactions are not bridged at all. And any subsequent Discord user which tries to react will simply be ignored. This specific emoji can never be seen on Matrix side. Having a Discord user use another emoji to react does work.

src/bot.ts
```diff
@@ -1177,6 +1192,146 @@ export class DiscordBot {
...
+
+        while (storeEvent.Next()) {
+            const matrixIds = storeEvent.MatrixId.split(";");
+
+            for (const room of rooms) {
+                const reactionEventId = await intent.underlyingClient.unstableApis.addReactionToEvent(
+                    room,
+                    matrixIds[0],
+                    reaction.emoji.id ? `:${reactionName}:` : reactionName
+                );
```
salixor: We stumbled on an issue on an hosted bridge : if there are multiple Matrix rooms linked to a Discord channel, we can get the following message : `Relations must be in the same room`
If it fails on a room, then further rooms are ignored. Maybe a `try/catch` would be a nice addition to ignore this issue.

src/bot.ts
SethFalco: You're right, we don't need rooms here. I misunderstood the contents of the `storeEvent` object. Will update this to use your implementation, which also resolves the bug reported by @salixor below.

src/bot.ts
SethFalco: I've rebased the branch with `develop` and resolved this.
Feel free to test it whenever you find time for it.

SethFalco:
@salixor @mangofeet I've rebased with `develop`, and resolved the exceptions when users aren't in the room when they add/remove reactions now.
I'd appreciate it if you could test it, just to confirm everything is good. :+1: 

I also found another bug. Removing all reactions in the Discord interface should bridge to Matrix correctly now.

salixor:
Hey there, haven't had much time last week to try the changes out, but I haven't forgotten !

I'm glad to say the first issue is solved !

The second issue is still a weird one.
The reactions have been bridged on one message, but not the other - with no apparent error.
On the first message, adding the same reaction myself removes the reaction of the Discord bot user on Matrix's side, as shown below :
image
image
It's a very niche use case however, so I'd say it works well enough. At least actual user reactions are always bridged, which is the most important imo.

For the multiple room issue, we only stumbled on this use case because of an old room which bridge was kinda jumbled, and I don't really know how to replicate the same use case - sorry ! Also a niche use case, and the changes seem reasonable.

Really hope it gets merged someday, and thanks a lot for the work !
Gonna run my bridge with these new changes, and I'll report if something weird happens.

PS : I tried the remove all feature but it doesn't seem to interface properly. I get no errors but the reaction stay on Matrix's side

GitMensch:
Is there anything open but the rebase (same question goes for the related #877)?
Who would handle the merge?

src/bot.ts
```diff
@@ -1177,6 +1192,151 @@ export class DiscordBot {
...
+                eventId,
+                "m.annotation"
+            );
+
+            const event = chunk.find((event) => {
+                if (event.sender !== intent.userId) {
+                    return false;
+                }
+
+                return event.content["m.relates_to"].key === reaction.emoji.name;
```
chayleaf, line 1283: you should use `reactionName` here in the same way as in `OnMessageReactionAdd`

test/test_discordbot.ts
```diff
@@ -442,4 +444,85 @@ describe("DiscordBot", () => {
...
+                "$mAKet_w5WYFCgh1WaHVOvyn9LJLbolFeuELTKVfm0Po",
+                "🤔"
+            );
+        });
+
+        it("Removes reaction from Discord → Matrix", async () => {
+            discordBot = getDiscordBot();
+            const intent = mockBridge.getIntent(author.id);
+
+            intent.underlyingClient.unstableApis.getRelationsForEvent = async () => {
```
chayleaf, line 504: this is not unstable anymore, using it via `.unstableApis` is deprecated

```suggestion
            intent.underlyingClient.getRelationsForEvent = async () => {
```

src/bot.ts
```diff
@@ -1177,6 +1192,151 @@ export class DiscordBot {
...
+            if (user.partial) {
+                log.warn(`Skipping reaction remove for user with Discord ID ${user.id} in ${roomId}. User was partial.`);
+                continue;
+            }
+
+            await this.userSync.JoinRoom(user, roomId);
+
+            const underlyingClient = intent.underlyingClient;
+
+            const { chunk } = await underlyingClient.unstableApis.getRelationsForEvent(
```
chayleaf, line 1272:
```suggestion
            const { chunk } = await underlyingClient.getRelationsForEvent(
```

src/bot.ts
```diff
@@ -1177,6 +1192,151 @@ export class DiscordBot {
...
+        if (!storeEvent?.Result) {
+            log.verbose(`Received remove all reaction event for untracked message. Dropping! Event: ${message}`);
+            return;
+        }
+
+        while (storeEvent.Next()) {
+            const [ eventId, roomId ] = storeEvent.MatrixId.split(";");
+            const underlyingClient = this.bridge.botIntent.underlyingClient;
+
+            const { chunk } = await underlyingClient.unstableApis.getRelationsForEvent(
```
chayleaf, line 1322:
```suggestion
            const { chunk } = await underlyingClient.getRelationsForEvent(
```

test/mocks/appservicemock.ts
```diff
@@ -276,4 +277,19 @@ class MatrixClientMock extends AppserviceMockBase {
...
+
+class UnstableApis extends AppserviceMockBase {
+
+    public async addReactionToEvent(roomId: string, eventId: string, emoji: string) {
+        this.funcCalled("addReactionToEvent", roomId, eventId, emoji);
+    }
+
+    public async getRelationsForEvent(roomId: string, eventId: string, relationType?: string, eventType?: string): Promise<any> {
+        this.funcCalled("getRelationsForEvent", roomId, eventId, relationType, eventType);
+    }
```
chayleaf, line 294: move to MatrixClientMock

chayleaf (changes requested):

SethFalco:
I've actually deleted my Discord account since creating this PR.
Would anyone care to take it off my hands to continue maintaining it?

Edit: I won't close it unless someone wants to take over. In future, I may create a dummy account just to update the PRs if no one else wants to maintain it.

PR #869, opened 2022-11-09
Fix matrix metrics handling

tadzik:
This fixes 3 separate crashes seemingly introduced in ed62d8add13137a425f818ec4704c3b58fa111c4 that somehow slipped through for all this time – presumably due to metrics being off by default.


Half-Shot (approved):
Ugh, how is our type checking *this* bad :(

PR #877, opened 2023-02-13
feat: add Matrix -> Discord reactions

mangofeet:
Implements reactions from matrix to discord, intended for cases where a single matrix user is using the bot as a personal puppet

Functionality is disabled by default and enable by setting `bridge.enableMatrixReactions` to `true`

- [x] Send reactions to discord from Matrix to Discord
- [x] Removes reactions when redacted in Matrix
- [x] Handle thumbs up exception from Matrix to Discord
- [ ] Handle thumbs up exception from Discord to Matrix

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,8 @@
 FROM node:16-slim AS BUILD
-COPY . /tmp/src
 # install some dependencies needed for the build process
 RUN apt update && apt install -y build-essential make gcc g++ python3 ca-certificates libc-dev wget git
 
+COPY . /tmp/src
 RUN cd /tmp/src \
     && yarn
 
--- a/config/config.sample.yaml
+++ b/config/config.sample.yaml
@@ -38,7 +38,7 @@ bridge:
   disableInviteNotifications: false
   # Disable Room Topic echos from matrix
   disableRoomTopicNotifications: false
-  # Enable reactions from matrix to discord. This will only respond to
+  # Enable reactions from Matrix to Discord. This will only respond to
   # the first reaction of each emoji and is intended for cases where
   # the bridge is used by a single matrix used
   enableMatrixReactions: false
--- a/config/config.sample.yaml
+++ b/config/config.sample.yaml
@@ -38,6 +38,10 @@ bridge:
   disableInviteNotifications: false
   # Disable Room Topic echos from matrix
   disableRoomTopicNotifications: false
+  # Enable reactions from matrix to discord. This will only respond to
+  # the first reaction of each emoji and is intended for cases where
+  # the bridge is used by a single matrix used
+  enableMatrixReactions: false
   # Auto-determine the language of code blocks (this can be CPU-intensive)
   determineCodeLanguage: false
   # MXID of an admin user that will be PMd if the bridge experiences problems. Optional
--- a/config/config.sample.yaml
+++ b/config/config.sample.yaml
@@ -40,7 +40,7 @@ bridge:
   disableRoomTopicNotifications: false
   # Enable reactions from Matrix to Discord. This will only respond to
   # the first reaction of each emoji and is intended for cases where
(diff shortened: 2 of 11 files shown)


Dockerfile
```diff
@@ -1,8 +1,8 @@
 FROM node:16-slim AS BUILD
-COPY . /tmp/src
 # install some dependencies needed for the build process
 RUN apt update && apt install -y build-essential make gcc g++ python3 ca-certificates libc-dev wget git
 
+COPY . /tmp/src
```
jaller94, line 5: Nice optimisation for having to build the first step less often. :+1:

src/bot.ts
```diff
@@ -680,6 +680,42 @@ export class DiscordBot {
...
+
+        const storeEvent = await this.store.Get(DbEvent, { matrix_id: `${relatesTo.event_id};${event.room_id}` });
+
+        if (!storeEvent || !storeEvent.Result) {
+            log.warn(`Could not react because the event was not in the store.`);
+            return;
+        }
+        log.info(`React event matched ${storeEvent.ResultCount} entries`);
+        while (storeEvent.Next()) {
+            log.info(`Reacting to discord msg ${storeEvent.DiscordId}`);
```
jaller94:
```suggestion
            log.info(`Reacting to Discord msg ${storeEvent.DiscordId}`);
```

config/config.sample.yaml
```diff
@@ -38,6 +38,10 @@ bridge:
   disableInviteNotifications: false
   # Disable Room Topic echos from matrix
   disableRoomTopicNotifications: false
+  # Enable reactions from matrix to discord. This will only respond to
```
jaller94: Capitalise names to differentiate them from the regular English words.

```suggestion
  # Enable reactions from Matrix to Discord. This will only respond to
```

jaller94 (commented):
Looking forward to Discord reactions. :clap: 
Do you have everything you need to finish the PR?

mangofeet:
> Looking forward to Discord reactions. clap Do you have everything you need to finish the PR?

I still need to figure out how to respond to the redaction of a reaction in Matrix. Unfortunately it looks like I may have to figure out a way to store the records differently so that the event ID of the reaction that comes back in `event.redacts` can be used to look up the original Matrix event or at least the original Discord message, and it needs to save the emoji used so that it can remove the correct one I believe. 

I'll be checking out the code that's in #862 to maybe pull some implementation details from there (hopefully without depdending on that PR since it's still in review I guess)

Dockerfile
mangofeet: Wasn't sure if this should go in it's own PR, I was going to revert it, but yeah, it does speed up building by quite a bit when trying to test changes on a remote machine instead of locally.

EDIT: There's a way to make it even faster by adding just the `package.json` and `yarn.lock` files, doing the yarn install, then adding the source code and doing the build separately

mangofeet:
OK, so I got reactions redacting from matrix to discord. Still need to figure out the edge case of the thumbs up, but otherwise it's working as I intended now.

mangofeet:
Handled the thumbs up exception from Matrix to Discord, but not the other way around yet, not sure if that's strictly necessary to merge this though. It may be better to get this *and* #862 merged, and then handle any emoji exceptions in a unified manner after the fact.

src/bot.ts
```diff
@@ -726,13 +726,13 @@ export class DiscordBot {
 
         const relatesTo = event.content['m.relates_to'];
 
+        let emoji = relatesTo.key
+        if (emoji.indexOf("👍") >=0  ) {
+            emoji = "👍"
+        }
```
TheTechmage, line 732: After reading this if statement, I'm assuming that this is related to the skin-tones? If so, you may want to do proper parsing of those, as more than thumbs up is affected. There's also country flags which use a similar concept.

PR #887, opened 2023-06-18
Discord unique usernames change

erkinalp:
Discord unique usernames return "0" as the discriminator

### DIFF
--- a/src/matrixroomhandler.ts
+++ b/src/matrixroomhandler.ts
@@ -182,8 +182,8 @@ export class MatrixRoomHandler {
                     regexp: "[A-Za-z0-9_\-]{2,100}",
                 },
                 discriminator: {
-                    placeholder: "1234",
-                    regexp: "[0-9]{4}",
+                    placeholder: "0",
+                    regexp: "[0-9]{1,4}",
                 },
                 guild_id: {
                     placeholder: "",


src/matrixroomhandler.ts
```diff
@@ -182,8 +182,8 @@ export class MatrixRoomHandler {
                     regexp: "[A-Za-z0-9_\-]{2,100}",
                 },
                 discriminator: {
-                    placeholder: "1234",
-                    regexp: "[0-9]{4}",
+                    placeholder: "0",
+                    regexp: "[0-9]{1,4}",
```
AndrewFerr, line 186: This regexp allows 2- and 3-digit values, when AIUI the discriminator can only be either a 4-digit value or 0.  If that's the case, a regexp of `0|[0-9]{4}` should do the trick.

src/matrixroomhandler.ts
```diff
@@ -182,8 +182,8 @@ export class MatrixRoomHandler {
                     regexp: "[A-Za-z0-9_\-]{2,100}",
                 },
                 discriminator: {
-                    placeholder: "1234",
-                    regexp: "[0-9]{4}",
+                    placeholder: "0",
```
AndrewFerr, line 185: Consider leaving the placeholder as `1234` to serve as a general example of a discriminator for a non-unique username.  (And once `0` becomes the only effective discriminator once Discord's username changes get fully rolled out, the bridge can stop caring about them entirely.)

AndrewFerr (changes requested):
Thanks for your contribution!  When you have a chance, please sign-off your commit & add a changelog file.

src/matrixroomhandler.ts
erkinalp: It should be treated as a string. According to reverse engineering work done for https://github.com/spacebarchar/server, Discord's client also treats it as a string; and Discord itself likely also stores it as a string. I have previously tested alphanumeric discriminators against the Discord client and it worked fine. Discord discriminators should really be treated as opaque strings, just like Matrix room IDs.

src/matrixroomhandler.ts
erkinalp: On some third party reimplementations such as spacebar (https://github.com/spacebarchat/server/pull/1049), the fact that discriminators exist is itself configurable.

src/matrixroomhandler.ts
erkinalp: Discord and Matrix implementations currently send fixed length values for normally created Matrix rooms (not with a manually overridden room ID) and normally created Discord users (not with a manually overridden discriminator), but this is not guaranteed in the API.

erkinalp: @AndrewFerr already signed-off, see the commit message

PR #888, opened 2023-06-18
Prefer server-level display names when available

TheStranjer:
Sometimes server-level displayName values don't propagate correctly onto Matrix. This fixes that. Apparently you can't get the user-level display name in DiscordJS yet except in the dev version, and you're on a fork of a fork of a fork of that so I'm just no messing with it.

Signed-off-by: `Bob Arctor <neetzsche@tutanota.com>`

### DIFF
--- a/changelog.d/888.bugfix
+++ b/changelog.d/888.bugfix
@@ -1,0 +1,1 @@
+Prefer server-level display names when available.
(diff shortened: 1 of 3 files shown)


AndrewFerr (approved):
It remains to be seen why server-level display name changes can get missed in the first place, but this should be a safe workaround.  Thanks for your contribution!

PR #898, opened 2023-09-08
Update better-discord.js

AndrewFerr:
This also uses its new package registry at [gitlab.com](https://gitlab.com/mx-puppet/discord/better-discord.js/-/packages/6501312).


Half-Shot (approved):
Looks good assuming tested

